PR #1, opened 2016-07-29
fix duplicate notification

blaulan:
use 'encounter_id' instead of 'id' to avoid duplicate notification.

### DIFF
--- a/alarms/alarm_manager.py
+++ b/alarms/alarm_manager.py
@@ -44,6 +44,8 @@ class Alarm_Manager:
 			name = pkmn_name(pkmn['pokemon_id'])
 			dissapear_time = datetime.utcfromtimestamp(pkmn['disappear_time']);
 			pkinfo = {
+				'id': pkmn['pokemon_id'],
+				'name': name,
 				'alert': pkmn_alert_text(name),
 				'gmaps_link': gmaps_link(pkmn['latitude'], pkmn['longitude']),
 				'time_text': pkmn_time_text(dissapear_time),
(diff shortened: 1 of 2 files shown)


kvangent:
I like the changes on Slack. Are the icons we are using up in Public Domain though? I want to make sure we aren't pulling bandwidth from someone else.

kvangent:
We can use the ones in PokemonGo-Map if we can find a reasonable host. 

Would you please rebase for me as well?

reactortrip:
I used your other alarm til the Niantic blockade and switched to a modified version of the dev with your code to add the alarm back in.  I edited my slack alarm to include emoji support by adding in ":pokemon-" + Pokemon name + ":".  So using icons isn't necessary, you can just have people add the custom emojis from this page to their slack.  Won't fix other services though.

https://github.com/Templarian/slack-emoji-pokemon

reactortrip:
Pic of my alarms in slack

http://i68.tinypic.com/mazalg.jpg

kvangent:
@reactortrip If you would like to make a tutorial for how to do that, I can include it in the wiki.

blaulan:
I've modified the code to use @reactortrip 's method for pokemon image.

kvangent:
@blaulan Will they still need to add their own icons?

blaulan:
@kvangent Yes, the instructions can be found from https://github.com/lambtron/emojipacks. However, I'm still trying to find a high solution pokemon icon pack. Those from PokemonGo-Map are in a single image, which can't be passed to Slack.

kvangent:
@blaulan What about the ones in static/icons?

blaulan:
@kvangent we can use those for now, although their resolution is low. I will include them in the next commit.

kvangent:
Looks good, thanks for this.

PR #11, opened 2016-07-30
Correctly convert disappear_time, fixes notifications not triggering

schneefux:


kvangent:
This does not correct the problem with timestamps. The problem is actually on the Map side. They are sending the UTC time in a native object, which it then interprets as local time. This fix interprets it also as local time, which will compound the problem for some users. The correct fix is listed on the reddit thread (and I will soon have it updated to the ReadMe)

PR #17, opened 2016-08-01
Start a worker thread to handle the alerts

wbowling:
As @kvangent suggested, a much better approach using a [Queue](https://docs.python.org/2/library/queue.html) to pass data to a running thread.

### DIFF
--- a/runwebhook.py
+++ b/runwebhook.py
@@ -3,6 +3,8 @@ import logging
 import os
 
 from flask import Flask, request
+from Queue import Queue
+from threading import Thread
 
 from alarms import config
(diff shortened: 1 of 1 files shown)


kvangent:
@wbowling Awesome, thank you for working on this. I wish I had looked earlier, I was in the process of doing this myself. However I have one question - will the python daemon service still work in windows?

kvangent:
@wbowling Take a look at the multi_threading branch and tell me what you think.

kvangent:
@wbowling I'm going to close this since it's implemented in a different way now, but once again thank you for taking the time to do it and please let me know if you want to help with other features in the future.

PR #31, opened 2016-08-03
Implement pushing to Pushbullet channel

hhu94:
If the user fills the channel field in alarms.json, the script will attempt to push to that channel instead of pushing to all the devices of the given account.

### DIFF
--- a/alarms/pushbullet_alarm.py
+++ b/alarms/pushbullet_alarm.py
@@ -13,6 +13,7 @@ class Pushbullet_Alarm(Alarm):
 		if settings['channel']:
 			try:
 				self.channel = get_channel(self.client, settings['channel'])
+                                log.info('Pushing to channel "' + settings['channel'] + '"')
 			except PushbulletError:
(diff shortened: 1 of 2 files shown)


alarms/pushbullet_alarm.py
```diff
@@ -22,5 +29,15 @@ def pokemon_alert(self, pkinfo):
...
+		else:
+			push = self.client.push_link(notification_text, gmaps_link, body=time_text)
+
+def get_channel(client, channel_tag):
+	req_channel = next((channel for channel in client.channels if channel.channel_tag == channel_tag), None)
+
+	if req_channel is None:
+		raise PushbulletError('No channel found with channel_tag "{}"'.format(channel_tag))
+
+	return req_channel
```
hhu94: Since pushbullet.py version 0.10.0 doesn't have the get_channel function, I had to copy it over from https://github.com/randomchars/pushbullet.py/pull/73/files

PR #50, opened 2016-08-07
Added Support for Pushover.net Service

DeeFox:
Added support for the pushover.net push-messaging service for Android & iOS and wanted to share it.


kvangent:
I like this, but I need to clean it up and add some error handling before I pull it into master.

PR #75, opened 2016-08-10
Create geofencing/polygon option to improve on location abilities

brettgus:
## Description
Adding a geofencing option.  The current version allows one to specify a distance.  However, this option allows a user to specify a polygon which the pokemon must be found in for an alert to go off.

Useful for a situation such as where your PokemonGo-Map is monitoring a larger city area and you only want to notify on a certain neighborhood.

Also, while trying to catch pokemon, some areas we travel to may be easier to get to than others.  Consider the following scenario: heading west, we may have to go cross a river by bridge that is unsafe or slow for pedestrian travel.  Traveling north and east is easy.  To the south is a shady industrial district that isn't safe to go to at night.  The current location monitoring only allows us to create a radius from our current point to monitor.  However, drawing a polygon lets you take into account these factors yourself and draw an exact area of where you'd like to receive notifications for

## Screenshots:

Well, this isn't a screenshot of the change, but of the net effect it has on monitoring:
Screenshot

Current version:
Current version

## How Has This Been Tested?
Tested on my Ubuntu 16.04 install

## Types of changes

## Checklist:

Use [a tool like this](https://codepen.io/bgus/full/grBdrr/) to generate coordinates

python runwebhook -H 0.0.0.0 -P 5000 -ply '(43.05072,-87.9502),(43.05286,-87.9217),(43.04746,-87.89046),(43.0442,-87.89149),(43.0265,-87.89801),(43.03466,-87.91655),(43.03391,-87.93269),(43.03516,-87.94814)'

### DIFF
--- a/alarms/utils.py
+++ b/alarms/utils.py
@@ -14,6 +14,7 @@ from glob import glob
 from datetime import datetime, timedelta
 from math import radians, sin, cos, atan2, sqrt
 from s2sphere import LatLng
+from ast import literal_eval
 
 #Local imports
 from . import config
@@ -36,6 +37,7 @@ def set_config(root_path):
 	parser.add_argument('-L', '--locale', help='Locale for Pokemon names: default en, check locale folder for more options', default='en')
 	parser.add_argument('-d', '--debug', help='Debug Mode', action='store_true')
 	parser.add_argument('-tf', '--time_fix', help='Apply time_fix offset to received times.', action='store_true')
+	parser.add_argument('-ply', '--polygon', help='Only notify if pokemon are within geofence/polygon')
 	parser.set_defaults(DEBUG=False)
 	
 	args = parser.parse_args()
@@ -171,3 +173,22 @@ def get_pos_by_name(location_name):
 def time_fix(t):
 	diff = datetime.utcnow() - datetime.utcfromtimestamp(time.mktime(datetime.utcnow().timetuple()))
 	return t+diff
+	
+#Return true if x,y points are inside polygon
+def point_in_poly(x,y):
+        poly = config['POLYGON']
+        n = len(poly)
+        inside = False
+
+        p1x,p1y = poly[0]
+        for i in range(n+1):
+                p2x,p2y = poly[i % n]
+                if y > min(p1y,p2y):
+                        if y <= max(p1y,p2y):
+                                if x <= max(p1x,p2x):
+                                        if p1y != p2y:
+                                                xints = (y-p1y)*(p2x-p1x)/(p2y-p1y)+p1x
+                                        if p1x == p2x or x <= xints:
+                                                inside = not inside
+                p1x,p1y = p2x,p2y
+        return inside
--- a/alarms/utils.py
+++ b/alarms/utils.py
@@ -180,15 +180,11 @@ def point_in_poly(x,y):
         n = len(poly)
         inside = False
 
-        p1x,p1y = poly[0]
-        for i in range(n+1):
-                p2x,p2y = poly[i % n]
-                if y > min(p1y,p2y):
-                        if y <= max(p1y,p2y):
-                                if x <= max(p1x,p2x):
-                                        if p1y != p2y:
-                                                xints = (y-p1y)*(p2x-p1x)/(p2y-p1y)+p1x
-                                        if p1x == p2x or x <= xints:
-                                                inside = not inside
-                p1x,p1y = p2x,p2y
-        return inside
+	p1x, p1y = poly[0]
+	for i in range(n + 1):
+	p2x, p2y = poly[i % n]
+	if y > min(p1y, p2y) and y <= max(p1y, p2y) and x <= max(p1x, p2x):
+	    if p1x == p2x or x <= (y - p1y) * (p2x - p1x) / (p2y - p1y) + p1x:
+	        inside = not inside
+	p1x, p1y = p2x, p2y
+	return inside
--- a/alarms/alarm_manager.py
+++ b/alarms/alarm_manager.py
@@ -97,6 +97,11 @@ class Alarm_Manager(Thread):
 			log.info(self.notify_list[pkmn_id])
 			return
 		
+		if config["POLYGON"] = false :
+                        if point_in_poly(lat,lng) == False:
+                                log.info(name + " ignored: outside geofence/polygon")
+                                return
+                                
 		#Check if the Pokemon has already expired
 		if dissapear_time < datetime.utcnow() :
 			log.info(name + " ignore: time_left has passed.")


alarms/utils.py
```diff
@@ -170,4 +172,23 @@ def get_pos_by_name(location_name):
...
+        n = len(poly)
+        inside = False
+
+        p1x,p1y = poly[0]
+        for i in range(n+1):
+                p2x,p2y = poly[i % n]
+                if y > min(p1y,p2y):
+                        if y <= max(p1y,p2y):
+                                if x <= max(p1x,p2x):
+                                        if p1y != p2y:
```
kvangent: Instead of nesting so many if statements, can condense down to one?

alarms/utils.py
```diff
@@ -36,6 +37,7 @@ def set_config(root_path):
 	parser.add_argument('-L', '--locale', help='Locale for Pokemon names: default en, check locale folder for more options', default='en')
 	parser.add_argument('-d', '--debug', help='Debug Mode', action='store_true')
 	parser.add_argument('-tf', '--time_fix', help='Apply time_fix offset to received times.', action='store_true')
+	parser.add_argument('-ply', '--polygon', help='Only notify if pokemon are within geofence/polygon')
```
kvangent: I like this command, but can we use it to reference a file full of coordinates instead? Seems cleaner then a massive command to launch.

alarms/utils.py
```diff
@@ -170,4 +172,23 @@ def get_pos_by_name(location_name):
 	
 def time_fix(t):
 	diff = datetime.utcnow() - datetime.utcfromtimestamp(time.mktime(datetime.utcnow().timetuple()))
-	return t+diff
\ No newline at end of file
+	return t+diff
+	
+#Return true if x,y points are inside polygon
+def point_in_poly(x,y):
```
kvangent: Can we have some more commenting explaining the method used to determine if the point is in the geofence?

kvangent:
I really like the idea, but it needs a little bit of polishing (and a rebase at this point)

Also do you have a tool you use to draw a geofence to give you coordinates?

I'll leave this open for now, let me know if you want to go over any specifics. I'm really interested in including this.

brettgus:
I'll work to clean up the code as discussed in your comments.  Agreed, it's rough code, really just wanted to see if there was interest before I put too much time into it.

I've been using [this basic codepen](https://codepen.io/bgus/full/grBdrr/) to draw coordinates.  I could make something more functional if needed, this was mostly just a POC if there isn't another tool out there to do this.

Agreed, the command line, especially for a large polygon, can get pretty unwieldy.  I assumed the reason you made the location a command line flag was for flexibility, which is why I followed suit like that.  Would you want it in the alarms.json file or file that is passed to the command line?

Also, how would it be dealt with if the user specifies a polygon and location?  Seems like there should be some checking to make it so that the user could only use one at a time?  Or only use the location for distance tracking if a polygon is specified?  Thoughts?

kvangent:
Maybe eventually if a frontend ever gets put up we can use the codepen in the app. For now I'm fine with just leaving the url in the wiki. 

I'm not too worried about them using a location and a polygon - if they ONLY want the geofence they just have to set all the pokemon to "True" instead of specifying a distance. I think the best way to do it would be `-gf, --geofence` and give it the name of the file we want to use as coords.

I was thinking about this some more after I posted earlier too. I think we can eliminate some of the math to speed up the calculation. If we think of the points as a series of line segments, and stretch the location in question out to a line, we should just need to count the number of segments that we hit. We can just evaluate each segment and if our y coordinate is between the y coordinates of the other points, then it is a hit.

Rough Psuedocode:
```
intersect = False
for each point P in coords:
    NP = next point
    low = min(P.y, NP.Y)
    high = min.max(P.y, NP.y)
    if low < loc.y and loc.y < high:
         intersect = !intersect
```

We could even do the low/high calc when we read in the config file to make it easier.

brettgus:
The algorithm is a raycasting algorithm.  It's fast and works... but there are some edge cases where it could become problematic.  If the polygon crosses the equator/international date line, that might cause an issue, haven't tested that.  Also, it doesn't account for the spherical nature of the earth, so if it was a huge or very complex polygon.  I was thinking I could find a library that includes an algorithm that is bulletproof, but the major drawback is that there may be computational overhead involved.

kvangent:
I finished cleaning this up some and pulled is in via the command line, but it doesn't seem to mark it down 

Added via commit 29cda58a215d7197a58bca8adcb31468042b6744 - Credit to @brettgus for getting it started.

paulharman:
Is this only available via the command line?  I would love to be able to set this in alarms.json so each alarm can have its own settings.

brettgus:
@paulharman At the command line you specify the file containing the coordinates: 

Specify a file of coordinates, limiting alerts to within this area

Use either -gf or --geofence followed by a file containing a list of coordinates

rrbailey89:
I get this error when running the command. 

`F:\Ray\downloads\pogomap\pokealarm [master ≡ +1 ~1 -0 !]> python runwebhook.py -gf geofence.txt
2016-08-18 08:48:03,891 [          geofence] [   INFO] got to here
  File "runwebhook.py", line 38, in <module>
    config = set_config(os.path.abspath(os.path.dirname(__file__)))
  File "F:\Ray\downloads\pogomap\pokealarm\alarms\utils.py", line 54, in set_config
    config['GEOFENCE'] = Geofence(os.path.join(root_path, args.geofence))
  File "F:\Ray\downloads\pogomap\pokealarm\alarms\geofence.py", line 20, in __init__
    p = Point(float(row[0]), float(row[1]), evaluate=False)
ValueError: could not convert string to float: (33.26681`

kvangent:
Your geofence.txt should be in lat,lon format, with no parenthesis around them.

rrbailey89:
@kvangent oh okay I thought that the way it was displayed at the link above was okay. I got it working though. Thank you.

rrbailey89:
@kvangent not sure if you saw my ninja edit.

kvangent:
I didn't... Are you still receiving the error?

rrbailey89:
@kvangent yes I am. I believe its triggered by an alarm being triggered.

kvangent:
@rrbailey89 What does your coordinates.txt look like?

rrbailey89:
This is the geofence file im trying to use @kvangent 

[geofence.txt](https://github.com/kvangent/PokeAlarm/files/427676/geofence.txt)

kvangent:
@rrbailey89 Ok give me some time to play around with this. Could you make sure you installed requirements.txt and see if you still have the issue?

rrbailey89:
@kvangent thanks for looking into it. After running `pip install -r requirements.text --upgrade` just to be sure. I then run the webhook with geofence and receive this error immediately after it loads everything 

`2016-08-19 12:40:44,592 [          geofence] [   INFO] Geofence successfully loaded!
2016-08-19 12:40:44,602 [     alarm_manager] [   INFO] You will be notified of the following pokemon:
Gyarados, Venusaur, Ditto, Charmeleon, Charizard, Wartortle, Blastoise, Ivysaur, Snorlax, Articuno, Zapdos, Moltres, Dratini, Dragonair, Dragonite, Mewtwo, Mew, Ninetales, Jigglypuff, Wigglytuff, Magnemite, Magneton, Seel, Dewgong, Grimer, Muk, Gastly, Haunter, Gengar, Voltorb, Electrode, Exeggutor, Lickitung, Koffing, Weezing, Chansey, Tangela
2016-08-19 12:40:44,608 [     alarm_manager] [   INFO] Alarm not activated: pushbullet because value not set to "True"
2016-08-19 12:40:45,460 [       slack_alarm] [   INFO] Slack Alarm intialized.
2016-08-19 12:40:45,707 [     alarm_manager] [   INFO] Alarm not activated: twilio because value not set to "True"
2016-08-19 12:40:45,707 [     alarm_manager] [   INFO] Alarm not activated: telegram because value not set to "True"
2016-08-19 12:40:45,709 [     alarm_manager] [   INFO] Alarm not activated: pushover because value not set to "True"
2016-08-19 12:40:45,710 [     alarm_manager] [   INFO] Alarm not activated: twitter because value not set to "True"
2016-08-19 12:40:45,711 [     alarm_manager] [   INFO] PokeAlarm has started! Your alarms should trigger now.
2016-08-19 12:40:45,713 [        runwebhook] [   INFO] Webhook server running on http://127.0.0.1:4000
2016-08-19 12:40:45,740 [     alarm_manager] [   INFO] Krabby ignored: notify not enabled.
2016-08-19 12:40:45,743 [     alarm_manager] [   INFO] Poliwag ignored: notify not enabled.
Exception in thread Thread-1:
  File "C:\Python27\lib\threading.py", line 801, in __bootstrap_inner
    self.run()
  File "F:\Ray\downloads\pogomap\pokealarm\alarms\alarm_manager.py", line 72, in run
    self.trigger_pkmn(data['message'])
  File "F:\Ray\downloads\pogomap\pokealarm\alarms\alarm_manager.py", line 109, in trigger_pkmn
    if 'GEOFENCE' in config and not config['GEOFENCE'].contains(lat,lng):
  File "F:\Ray\downloads\pogomap\pokealarm\alarms\geofence.py", line 37, in contains
    return self.polygon.encloses_point(Point(x,y))
AttributeError: 'Point' object has no attribute 'encloses_point'`

kvangent:
@rrbailey89 I think the problem was there was an empty line at the end of your file that might have been mucking it up. I've made the input a little more robust. Try a new pull and let me know if that fixes it.

rrbailey89:
@kvangent I did a new pull and it ran longer but in the end still the same error (i removed that empty line also)

```
Exception in thread Thread-1:
  File "C:\Python27\lib\threading.py", line 801, in __bootstrap_inner
    self.run()
  File "F:\Ray\downloads\pogomap\pokealarm\alarms\alarm_manager.py", line 72, in run
    self.trigger_pkmn(data['message'])
  File "F:\Ray\downloads\pogomap\pokealarm\alarms\alarm_manager.py", line 109, in trigger_pkmn
    if 'GEOFENCE' in config and not config['GEOFENCE'].contains(lat,lng):
  File "F:\Ray\downloads\pogomap\pokealarm\alarms\geofence.py", line 40, in contains
    return self.polygon.encloses_point(Point(x,y))
AttributeError: 'Point' object has no attribute 'encloses_point'
```


Is there some way that I can provide you with debug information?

kvangent:
@rrbailey89 Can you open an issue ticket with this? Make sure to fill out the python and OS version.

Also trying running with `--debug` flag and see if you get anything new.

rrbailey89:
@kvangent I have created the issue.

PR #80, opened 2016-08-10
Added Postal Code to addr

RiberoD07:
<addr> used to give housenumber street, now it'll give housenumber street postal
## Description

looked into geocoder and saw that 'postal' was the field for the postal code, simply added it to addr via loc.postal which is parsing it from geocoder.google

## How Has This Been Tested?

added it to utils.py and tested it by adding addr to what is printed in my notifications for telegram, pushbullet, and twilio all worked as expected

Windows 10 x64
Samsung Galaxy s6 edge
used Twilio, Pushbullet and Telegram to test results

Changes don't affect other areas of the code negatively, a simple addition to addr which makes it provide more information to the user as requested
## Types of changes

## Checklist:

### DIFF
--- a/alarms/utils.py
+++ b/alarms/utils.py
@@ -91,7 +91,7 @@ def get_pkmn_name(pokemon_id):
 #Returns a String representing the nearest address to a lng, lat	
 def get_address(lat, lng):
 	loc = geocoder.google([lat,lng], method='reverse')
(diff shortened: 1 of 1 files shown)


kvangent:
@RiberoD07 I would prefer that this feature be added as a dynamic text substitution so that users can decide if they want to use it in the address or not. Feel free to add it in and I'll accept as a new PR.

PR #84, opened 2016-08-10
Added Dockerfile

annessi:
Added a basic Dockerfile to run PokeAlarm inside a docker. 
PokemonGo-Map needs to be running inside another docker.

## How Has This Been Tested?
Tested on Synology

## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@
 #   docker build -t pokealarm
 #   docker run -d --net container:PokemonGo-Map --name PokeAlarm -P pokealarm
 # Change "PokemonGo-Map" to the name of your PokemonGo-Map docker
+# For newer versions of docker maybe you have to change --net to --network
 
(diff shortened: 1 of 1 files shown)


kvangent:
So the reason I'm leaving this open is because I'm going to switch all the Alarms over to self-contained modules that lazy import - They won't import the needed external modules (like Slacker or Twitter) until they are actually created and if not installed will install via pip. This will most likely not work with this docker.

If you can test and confirm that this is not the case, or create a suitable workaround I'll be glad to accept this.

annessi:
As soon as you switch I'll test the dockerfile.

kvangent:
@annessi I've switched the remaining packages over.

annessi:
I've tested the Dockerfile and it's working just fine also with the modules.

screen shot 2016-08-18 at 08 46 23

PR #109, opened 2016-08-13
Added travel time to Pokemon

brettgus:
## Description
Allows for addition of &lt;walking_time&gt; and &lt;driving_time&gt; variables when location is set.  Will estimate walking/driving time to reach pokemon. 

## How Has This Been Tested?
Tested with multiple different locations and several edge cases (between multiple continents).
Ran on Ubuntu 16.04 with python 2.7
However, 

## Types of changes

## Checklist:

### DIFF
--- a/alarms/alarm_manager.py
+++ b/alarms/alarm_manager.py
@@ -125,8 +125,9 @@ class Alarm_Manager(Thread):
 			'time_left': timestamps[0],
 			'12h_time': timestamps[1],
 			'24h_time': timestamps[2],
-			'dir': get_dir(lat,lng)
-			
+			'dir': get_dir(lat,lng),
+			'walking_time': "%s" % get_travel_time([lat, lng], "walking"),
+			'driving_time': "%s" % get_travel_time([lat, lng], "driving")
 		}
 		
 		for alarm in self.alarms:
--- a/alarms/utils.py
+++ b/alarms/utils.py
@@ -14,6 +14,8 @@ from glob import glob
 from datetime import datetime, timedelta
 from math import radians, sin, cos, atan2, sqrt
 from s2sphere import LatLng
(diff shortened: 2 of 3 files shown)


fgnn:
Thanks for adding this. Cleaner than my implementation in #113. Some code from both can probably be merged.


Missing features compared to #113 are 1) a go-nogo icon being available based on walking distance,  2) bearing and compass direction arrows, 3) API key added (although we should never get to 2500 notifications per day, a dev may though. It should probably be added as optional)

brettgus:
Why add the API key option though? If google doesn't require it, I don't see why you'd need to add it.  Why count usage against your API key if you don't need to?  

The only reason to add it would be if we needed to get transit directions, as then it is required.... so I guess that could be a compelling reason.  Although is anyone using transit to catch pokemon?  I would think the amount of wait time required (for the train/bus) would not make it worth doing.

And yeah, agreed, there's definitely stuff that can be merged at first glance.  I'm usually driving, so compass direction wasn't even a thought to me, but that could definitely be helpful.  Also, checking whether you'll make it there before expire time is helpful.

brettgus:
Ahh @mrchops1024 just mentioned this in the other PR... that they throttle you for a little while if you request too too frequently without an API key.

brettgus:
@kvangent If there's interest in pulling this into the project, I could integrate elements of @fgnn's code in #113 into this.

If so, would also like to hear your opinion on whether the API key should be added.

kvangent:
@brettgus Not that there is not interest, I just had a long weekend and am in the process of getting things caught back up. I need some time to look them over and figure out differences before I decide.

kvangent:
Close because in need of a rebase. Hoping to implement with google api rework.

PR #113, opened 2016-08-13
Google maps API integration for walking time/distance

fgnn:
## Description
For all alerts that a notification is sent for, a query to google maps is done. We retrieve the walking time and distance and make it an available field for the alerts.

Google maps api key has been added to the json config file. Use the same one as in PokemonGo-Map.

Some extra calculations are thrown in:

An indicator as to whether or not you can walk to the pokemon in time to catch it
A unicode arrow indicating the compass direction you must walk. 8 unicode arrows are available.

## How Has This Been Tested?
Been running on my own system for several days. Windows.

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Wiki Update
1) Add your google maps api key to the json file.
2) Adjust your messages as necessary. See slack for an example. It is put in username since my phone only gets two lines on the notifications.

## Checklist:

- [x ] My code follows the code style of this project.
- [x ] My change requires a change to the documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 *.pyc
 alarms.json
 coords.txt
+/.project
+/.pydevproject
--- a/alarms.json.default
+++ b/alarms.json.default
@@ -40,6 +40,7 @@
 			"consumer_secret": "YOUR_CONSUMER_SECRET"
         }
 	],
+	"gmaps-key": "",
 	"pokemon":{
 		"Bulbasaur":"False",
 		"Ivysaur":"False",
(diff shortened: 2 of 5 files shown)


brettgus:
Ha beat you to the punch, ended up doing a PR for my implementation as well #109.  Can you get away with not using an API key with this gmaps package?  Google maps doesn't require it for distance matrix, but it wouldn't be that hard to hit the 2500 request limit if you're using it every time you notify regarding a pokemon.  
https://developers.google.com/maps/documentation/distance-matrix/usage-limits

fgnn:
Guess I was too slow :) 

Depenfing how many pokemon you have filtered for

mrchops1024:
There's also a throttle limit when using their API with no key. Too many requests too frequently will cause them to reject requests for a bit.

kvangent:
Closing this since in need of a rebase. Hoping to implement this with the google api rework.

PR #127, opened 2016-08-17
Fix Geofence

gkuiper:
## Description

The geofence option didn't work for me. Locations that were inside the geofence didn't trigger an alarm. 

In my implementation I use the SymPy.geomery library to do the calculation if a location is inside or outside the geofence. Two options are implemented to specify a geofence using 2 or more coordinated. When 2 are given a rectangle is created, otherwise a polygon is made.

## How Has This Been Tested?

Tested it on my own Ubuntu 14.04 system and using Docker. It is running for a day already.

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ X] My code follows the code style of this project.

### DIFF
--- a/alarms/geofence.py
+++ b/alarms/geofence.py
@@ -5,66 +5,35 @@ log = logging.getLogger(__name__)
 #Python Modules
 import sys
 import csv
-from collections import namedtuple
-
-Point = namedtuple('Point', 'x, y')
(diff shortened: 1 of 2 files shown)


kvangent:
Look good. Much simpler than the original,

PR #140, opened 2016-08-21
[Enhancement] Add switch to disable first tweet at initialization

Becold:
## Description
Add optional setting `startup_message` for Twitter module.
Default on "True".
If set to "False", it prevent the Alarm from tweeting "PokeAlarm has intialized!" at startup

## How Has This Been Tested?
With my twitter, 2 times and it work.

image

image

## Types of changes

## Wiki Update
Add to the optionnal options something like : 
```
`startup_message` | `Trigger a tweet at initialization` | `True`
```
## Checklist:

### DIFF
--- a/alarms/Twitter/twitter_alarm.py
+++ b/alarms/Twitter/twitter_alarm.py
@@ -21,9 +21,11 @@ class Twitter_Alarm(Alarm):
 		self.con_secret = settings['consumer_key']
 		self.con_secret_key = settings['consumer_secret']
 		self.status = settings.get('status', "A wild <pkmn> has appeared! Available until <24h_time> (<time_left>). <gmaps>")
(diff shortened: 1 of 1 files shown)


Becold:
Wait. Dont merge it.
It broke

gkuiper:
Wouldn't it be better to make this a global option? So also disable the initial message for the other services?

kvangent:
@gkuiper I can add this feature in to the rest of the alarms if there is interest. 

@Becold There is a parse Boolean function in the utils designed to interpret a number of things as True. You should make use of it here. 

I also think that you should change the name of the parameter to 'startup_message'.

kvangent:
One more thing that I just noticed that that you have the time as `utcnow()` instead of just `now()`. This will reintroduce the incorrect timezone you wanted changed before as well.

PR #171, opened 2016-08-27
Support sending notifications in a specified timezone.

ravishivt:
## Description

This adds a new CLI option, -tz or --timezone, that sends notifications in that timezone instead of always using UTC formatted dates.  If an option isn't specified, it defaults to "UTC".  A list of valid timezone strings are available at https://en.wikipedia.org/wiki/List_of_tz_database_time_zones.

One related minor changes is adding seconds to time_12 to be consistent with time_24.

## How Has This Been Tested?

Tested specifying a timezone and tested again without.
python runwebhook.py -H 0.0.0.0 -P 4002 --config alarms.TAMARACK.json -k "XXXX" -tz "America/Los_Angeles"

image

## Types of changes

## Checklist:

- [] I have updated the documentation accordingly.

### DIFF
--- a/alarms/utils.py
+++ b/alarms/utils.py
@@ -215,7 +215,7 @@ def get_timestamps(t):
 	disappear_time = disappear_time.replace(tzinfo=pytz.utc).astimezone(config.get("TIMEZONE"))
 	time_left = "%dm %ds" % (m, s)
 	time_12 = disappear_time.strftime("%I:%M:%S") + disappear_time.strftime("%p").lower() + disappear_time.strftime(" %Z")
-	time_24 = disappear_time.strftime("%H:%M:%S%Z")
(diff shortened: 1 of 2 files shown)


kvangent:
Currently it sends the time in local time of the machine it is set up on. I think that this would be a better idea when defaulting to UTC time.

ravishivt:
I always set servers to UTC time as I've learned it is a best practice.  Good discussion on this at http://stackoverflow.com/questions/2532729/daylight-saving-time-and-time-zone-best-practices

Regardless, I've modified the PR so it plays nicely with the machine's local timezone.  The timezone option doesn't default to anything and it only translates the timestamps if the machine's local timezone is different than the requested timezone option.

kvangent:
@ravishivt I think it looks good, We're about to merge a big change, and so I'll reevaluate it after that and see if it need a rebase or not.

PR #203, opened 2016-09-05
Feature: Set an alarm to be triggered only within a given area (geopoint + radius)

PescaCode:
## Description

- **Feature**:  Set an alarm to be triggered only within a given area (geopoint + radius). This allows setting different alarms for different areas.
- **Minor fix**: Server will accept webhooks from all IPs (0.0.0.0)
- **Minor fix**: Default value (from a none existing settings) are now Bool type (False instead of "False")
- **Styling**: Some blank lines have been removed for better styling.

## How Has This Been Tested?

- Tested with geo distance of 200m, using Twitter. Works well.

## Types of changes

## Wiki Update

To set an alarm to be triggered only in a given area, it is possible to set a geo-location and radius in the alarm.json file as following:

```
	"alarms":[{
		    "active": "True",
		    "type": "twitter",
		    "access_token": "ACCESS_TOCKEN",
		    "access_secret": "ACCESS_SECRET",
		    "consumer_key": "CONSUMER_KEY",
		    "consumer_secret": "CONSUMER_SECRET",
		    "startup_message": "False",
		    "startup_list": "False",
		    "location": [00.00000, 00.00000],
		    "distance": 200
	}],
```
## Checklist:

- [x] I have updated the documentation accordingly.

### DIFF
--- a/alarms.json.default
+++ b/alarms.json.default
@@ -42,10 +42,10 @@
 			"access_secret": "YOUR_ACCESS_SECRET",
 			"consumer_key": "YOUR_CONSUMER_KEY",
 			"consumer_secret": "YOUR_CONSUMER_SECRET",
-		    "startup_message": "Optional: True/False",
-		    "startup_list": "Optional: True/False",
-		    "location": [00, 00],
-		    "distance": 0
+			"startup_message": "Optional: True/False",
+			"startup_list": "Optional: True/False",
+			"location": [00, 00],
+			"distance": 0
         }
 	],
 	"gyms":{
--- a/alarms.json.default
+++ b/alarms.json.default
@@ -41,7 +41,11 @@
 			"access_token": "YOUR_ACCESS_TOKEN",
 			"access_secret": "YOUR_ACCESS_SECRET",
 			"consumer_key": "YOUR_CONSUMER_KEY",
-			"consumer_secret": "YOUR_CONSUMER_SECRET"
+			"consumer_secret": "YOUR_CONSUMER_SECRET",
+		    "startup_message": "Optional: True/False",
+		    "startup_list": "Optional: True/False",
+		    "location": [00, 00],
+		    "distance": 0
         }
 	],
 	"gyms":{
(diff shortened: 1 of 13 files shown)


kvangent:
I appreciate you taking the time to submit this and for filling out the template thoroughly. However there are a few issues with your submission:
> Feature: Set an alarm to be triggered only within a given area (geopoint + radius). This allows setting different alarms for different areas.

It is a conscious effort in the design to have one 'limitations' per alarm grouping. This is intentional because it simplifies the config process. Otherwise the customization per alarm will never end - next people will want to limit pokemon, then based on a geofence and eventually we will have people with complex customization for one alarm. Instead of going down that rabbit hole, we just limit ourselves to one set of limitations per grouping of alarms.

> Minor fix: Server will accept webhooks from all IPs (0.0.0.0) 

This setting could already be provided by specifing 0.0.0.0 via the HOST parameter. This change actually removes the ability to set the server up to listen only locally or to a specific IP instead of several.

> Minor fix: Default value (from a none existing settings) are now Bool type (False instead of "False")

The reason they were Strings was because we use a parseBoolean function to try and accept multiple Strings as `True` or `False` - so when we use `parseBoolean` on a real Boolean, it'll attempt a call on your Boolean.lower() (which probably won't work). On top of that, while the `string == "True"` looks like something that doesn't make sense, it was done to keep Strings defaulting to true. So if I were to specify a string "gazorpazorp"* instead of a value as True or False, it wont turn the alarm on. We would rather have a misspelling turn the alarm off (in which it's absence would be a red flag something is wrong).

*(I may be misunderstanding this as I'm pretty new to python, but I believe the statement `if string_val` evaluates to true unless the value is nonexistent, None, or some variation of "false" in Python)

Anyway, once again I appreciate you taking the time to submit this PR - it's always cool to see people wanting to contribute. However, the issues above make your additions a poor fit for our current project plan. (I am planning on implementing some multi core features in the future that would allow someone to run multiple configs with 1 webhook listener, but these changes are in the planning phase only and likely won't be done for a few weeks).

PR #214, opened 2016-09-09
support for different spawnpoint types & setup file

seikur0:
## Description
It'll enable third party tools to optionally include a new field named addinfo in the webhook request. That field brings support for all different spawn point types, 2x15,1x60h2,1x60h3 and 1x60h23.

An example: Say a 2x15 point is scanned right at the start, so without this pull request, you can either give an alarm for an expire time of 15 min, or 45 min. In reality the pokemon will stay for 15 min, hide for 15 min and then be back for 15 min, which would be a very useful information for the people receiving the alert, but it's not possible right now.

That addinfo is completely optional and for those not using it nothing changes.

This pull request changes that.

Oh yeah and I also added a setup file, which enables other projects to use PokeAlarm via "-e git+git://github.com/kvangent/PokeAlarm#egg=pokealarm" in their requirements.txt file.

## How Has This Been Tested?
I tested it in my head :D It's just some changed messages and unless this pull request gets approved, I won't integrate PokeAlarm into my own tool, so I haven't tested it.

## Types of changes

## Wiki Update
If you integrate it, I can make a short wiki page.

## Checklist:
- [x] My code follows the code style of this project. Tabs as indents, really? :/
- [x] My change requires a change to the documentation. Only if you want to advertise this feature I guess ;)


kvangent:
1. I like the idea of setup.py and don't have any issues with it. (But I haven't tested it yet.)

2. I'm like the idea of the 2x15 information, but I'm not keen on the implementation. It should really be a function in utils to match the others, and one that doesn't get called if the field isn't in the POST request.

3. I don't like that you added it to the defaults for each alarm. I would rather have it be an optional field so that people who use it consciously turn it on and then will see if fail to be replaced if something is working incorrectly. 

Do you have a link to the project that uses this in their webhooks?

seikur0:
2. It's not a very big thing, but I can understand why you'd rather have it in the utils, If the field isn't in the post request, it won't change anything, since it'll be just an empty string then, but again that's up to you.

3. If the addinfo field isn't in the request, the addinfo text will be just an empty string and change nothing about the messages at all, so I think it's good in the default messages.

I can make a branch of my project using this tomorrow, right now I'm using this addinfo system for my alerts and map as well, see https://github.com/seikur0/PGO-mapscan-opt/blob/fdb2bb08b04f2b435b0dbc09fdc9db310b466d46/main0.py around line 1300.

toshioson:
Looking for this feature for a while, fabulous! 

@seikur0 One thing I do not fully understanding from the coding hope you can explain. 

I can see that the trigger on reminder of these special spawnpoints (which lead to extra time of the Pokemon than the usual 15 mins) are based on 'addinfo'. Is 'addinfo' an existing data from pkmn in the existing develop branch of PokemonGo-Map? Or am I missing anything? Thanks in advance and sorry for being not ignorant in coding.

seikur0:
I don't think they do that or can even recognize these spawn point types properly, so it's not for PokemonGo-Map but for other mappers wanting to use PokeAlarm like me. For PokemonGo-Map users nothing will change, unless they integrate this as well.

(@kvangent I haven't forgotten it, I'm currently working on a bigger update and haven't had the time yet, but it'll be in that update as well.)

seikur0:
@kvangent: Get this [branch](https://github.com/seikur0/PGO-mapscan-opt/tree/pokealarm-demo) of my tool to test this pull request with.

It's working as intended and sending the hooks to localhost:4000.

Scan an area of your choice with some 2x15 points or other rare types in it, after 1 hour it'll create the learning file and enter iscan mode. And at the same time it knows the spawnpoint types, so it'll start sending addinfo values other than 0 and you'll notice the messages, I've added.

I guess you can also put line 1659 and 1660 in main0.py into an if addinfo: statement, so it'll only send the addinfo ones for testing purposes.

alarms/Boxcar/boxcar_alarm.py
```diff
@@ -17,7 +17,7 @@ class Boxcar_Alarm(Alarm):
 	_defaults = {
 		'pokemon':{
 			'title':"A wild <pkmn> has appeared!",
-			'long_message':"Available <a href='<gmaps>'>here</a> until <24h_time> (<time_left>).",
```
kvangent: I don't think it's necessary to edit the defaults. Since it's optional information, allow users to add this option in themselves with dynamic text substitution. Same for all the alarm files.

alarms/alarm_manager.py
```diff
@@ -172,7 +172,14 @@ def trigger_pokemon(self, pkmn):
 			'dir': get_dir(lat,lng)
 		}
 		pkmn_info = self.optional_arguments(pkmn_info)
-			
+
+		#Check if the Pokemon will be back after a hidden phase
+		#Addinfo values: 1=2x15 point, 2=1x60h2 point, 3=1x60h3 point, 4=1x60h23 point, only needs to be added, if scanned before the break
+		addinfo = pkmn.get('addinfo',0)
```
kvangent: This all should be wrapped in a function in utils. Ideally we would just add one parameter into the pokemon info that calls this function to get the fields value.

setup.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python
```
kvangent: I'm fine with this, but perhaps it should be in a separate PR since it's really separate from anything else.

alarms/Boxcar/boxcar_alarm.py
seikur0: I don't think it's any more optional than the time_left or the 24h_time one. Even less so, because those two are redundant. Do you really think it'll be a typical situation, that users don't want to know, that a pokemon will return after some time and how long? I don't think so, the realistic situation would be for them to never realize this feature exists at all. And for these reasons I think it has to be in the standard values.

alarms/alarm_manager.py
seikur0: I can do that.

setup.py
seikur0: And that as well.

alarms/Boxcar/boxcar_alarm.py
kvangent: I think it needs to be nonstandard because:
1. We primarily support PokemonGo-Map.
2. PokemonGo-Map doesn't support this feature.

I understand the way you have it now replaces it with empty space, but it should function similar to the other dynamic text options and instead leave a 'None' if the information was not provided in the POST request. If you are using a scanner with this function, it makes sense that you need to let PokeAlarm know that you want to see it. Otherwise it should be ignored.

seikur0:
Changed ans split into 2 pull requests.

PR #230, opened 2016-09-21
support for different spawnpoint types

seikur0:
## Description
It'll enable third party tools to optionally include a new field named addinfo in the webhook request. That field brings support for all different spawn point types, 2x15,1x60h2,1x60h3 and 1x60h23.

An example: Say a 2x15 point is scanned right at the start, so without this pull request, you can either give an alarm for an expire time of 15 min, or 45 min. In reality the pokemon will stay for 15 min, hide for 15 min and then be back for 15 min, which would be a very useful information for the people receiving the alert, but it's not possible right now.

This pull request changes that.

## How Has This Been Tested?
Tested with PGO-mapscan-opt

## Types of changes

## Wiki Update
If you integrate it, I can make a short wiki page.

## Checklist:
- [x] My change requires a change to the documentation. Only if you want to advertise this feature I guess ;)

### DIFF
--- a/alarms/alarm_manager.py
+++ b/alarms/alarm_manager.py
@@ -173,14 +173,6 @@ class Alarm_Manager(Thread):
 			'addinfo': get_addinfo(pkmn.get('addinfo',0))
 		}
 		pkmn_info = self.optional_arguments(pkmn_info)
-
-		#Check if the Pokemon will be back after a hidden phase
-		#Addinfo values: 1=2x15 point, 2=1x60h2 point, 3=1x60h3 point, 4=1x60h23 point, only needs to be added, if scanned before the break
-		addinfo = pkmn.get('addinfo',0)
-		reappear_texts = ('','\n15m later back for 15m.', '\n15m later back for 30m.', '\n30m later back for 15m.')
-		reappear_ind = (0, 1, 2, 1, 3)
-		pkmn_info['addinfo'] = reappear_texts[reappear_ind[addinfo]]
-
 		for alarm in self.alarms:
 			alarm.pokemon_alert(pkmn_info)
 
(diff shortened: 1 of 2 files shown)


alarms/alarm_manager.py
```diff
@@ -169,10 +169,10 @@ def trigger_pokemon(self, pkmn):
 			'time_left': timestamps[0],
 			'12h_time': timestamps[1],
 			'24h_time': timestamps[2],
-			'dir': get_dir(lat,lng)
+			'dir': get_dir(lat,lng),
+			'addinfo': get_addinfo(pkmn.get('addinfo',0))
```
kvangent: Can we change this to spawn_desc or something over addinfo to be more concise?

alarms/utils.py
```diff
@@ -278,6 +278,13 @@ def get_timestamps(t):
 	time_24 = disappear_time.strftime("%H:%M:%S")
 	return (time_left, time_12, time_24)
 	
+# Check if the Pokemon will be back after a hidden phase
+# Addinfo values: 1=2x15 point, 2=1x60h2 point, 3=1x60h3 point, 4=1x60h23 point, only needs to be added, if scanned before the break
+def get_addinfo(addinfo):
```
kvangent: Same here - rename to get_spawn_text or something similiar - otherwise it might be mistaken as 'address info' or something.

alarms/alarm_manager.py
seikur0: Okay I changed the names to respawn_info and respawn_text

seikur0:
One more thing, could you add the <respawn_text> to your [wiki](https://github.com/kvangent/PokeAlarm/wiki/Dynamic-Text-Subsitution)? :)

brettgus:
@seikur0 If you want to write it up you can do a PR on the wiki Github https://github.com/kvangent/PokeAlarmWiki

seikur0:
Wait how does that even work, it's like github-ception :D Okay, I'll do that

PR #239, opened 2016-09-25
Preserve old_team on gym webhooks

jnovack:
## Description

Preserve `old_team` when migrating to neutral so that there is only one announcement.  Next scan will properly announce:

    _old_team_ was defeated by _new_team_

Rather than two announcements:

    _old_team_ was defeated by Neutral
    Neutral was defeated by _new_team_

## How Has This Been Tested?

What's this word "tested"?

Not appropriate.

## Types of changes
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Wiki Update

Gym announcements on defeats will take another scan.  This has pros and cons.  The good is that there will be enough time to get a good sense of who dropped in.  The bad is that you must wait another full rotation of your scanner to send the announcement.

This will be moot once prestige calculation is merged, because then you can see "_GYM_ is being attacked..." as the prestige lowers.

## Checklist:

Will update documentation upon acceptance of idea.

### DIFF
--- a/alarms/alarm_manager.py
+++ b/alarms/alarm_manager.py
@@ -249,6 +249,13 @@ class Alarm_Manager(Thread):
 			log.debug("Gym ignored: no change detected")
 			return #ignore neutral for now
 		
+		# Preserve old_team so that there is no "Neutral" announcement
+		# Has the side effect of delaying the takedown until next scan.
+		if old_team == 0:
+			self.gyms[id] = old_team
+			log.debug("Gym ignored: migration to neutral, preserving old_team")
+			return
+		
 		#Check for Alert settings
 		old_team = get_team_name(old_team)
 		new_team = get_team_name(new_team)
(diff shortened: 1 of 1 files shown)


kvangent:
Just a heads up - I'm not ignoring these, I'm just waiting to deal with them until the IV/Moveset filtering is complete.

jnovack:
No problem.  I figured we would discuss this in discord when you had the
time.

kvangent:
So I am okay with this, but we need to find some way to make it optional. We could do a console command, or we could do a "Ignore_Neutral" in the Alarms.json. I am open to suggestions on the best way to handle this.

jnovack:
Hit +1 if you like it.

`alarms.json.default`:
```js
{
  "gyms": {
    "Ignore_Neutral": "False" /* "True" preserves the old_team when moving through Neutral */
  }
}
```

Sidenote, I HATE that you aren't using booleans.

kvangent:
@jnovack I'm fine with that setup. And I do use booleans. I parse the strings using the parse_booleans function from utils.py. This gives greater flexibility to what gets counted as True.

kvangent:
Feel free to reopen then when it's ready.

PR #290, opened 2016-12-23
Option to 'at channel' on slack for perfect pokemon

bellowsj:
## Description

Needed a way to be able to notify the channel when a 100% IV pokemon was found without double posting. The only way to do push notification in slack is using @channel (<!channel>) and currently there is no way to post to slack for 50% or greater of a certain Pokemon but do an @channel for 100% IV of that Pokemon without running two alarms and double posting. This feature automatically allows you to @channel the channel anytime a 100% IV pokemon is found. The slack section of the alarms.json now allow you to include `at_channel_on_perfect` and set that to True or False. If not specified it will default to false. If set to true then it will automatically add <!channel> to the body of any message if the Pokemon has 100% perfect IVs. 

## How Has This Been Tested?

Ran locally for several days. Tested using Charles proxy to modify the individual values of Pokemon to simulate a 100% IV of different Pokemon. 

Ran locally on my mac environment and also tested on my Ubuntu VPS.

## Types of changes

## Checklist:

### DIFF
--- a/alarms.json.default
+++ b/alarms.json.default
@@ -19,7 +19,8 @@
 		{
 			"active": "False",
 			"type":"slack",
-			"api_key":"YOUR_API_KEY"
+			"api_key":"YOUR_API_KEY",
+            "at_channel_on_perfect":"False"
 		},
 		{
 			"active": "False",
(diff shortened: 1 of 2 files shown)


kvangent:
I've looked this over and decided that to prevent feature bloat I am going to reject it. A much more general solution would be to just add an additional instance with the IV's min set to 100.

bellowsj:
@kvangent - Adding an additional instance with the min IVs set to 100 would result in double posting of the Pokemon, one with the <!channel> and one without. It would be confusing to users why it got posted twice.

PR #302, opened 2017-02-04
Include IV and Moves in Telegram Notifications

ashleykleynhans:
PokeAlarm supports IV and moves, but the telegram notifications don't show them.

This Pull Request includes the IV and moves in Telegram notifications.


kvangent:
This feature is already possible by customizing the Telegram message and taking advantage of Dynamic text substitution. I suggest you take a look at the wiki pages for implementations.

PR #313, opened 2017-02-17
Add new icons from Gen2

n30nl1ght:
Added gen2 images on icons dir.

image

Works fine with discord :-)


m3g4mInD:
How can we atache this icons on notifications?thanks for the great work.

PR #327, opened 2017-02-22
Ensure tweet length is less than max allowed

tomballgithub:
Tweets currently are allowed to be a maximum of 140 characters.  There
is no error checking in PokeAlarm at this time.  After text
substitutions, some tweets may be too long and this is not easily
noticed unless logs are monitored.  This PR avoids missing tweets.

## Description

This ensure that tweets being sent adhere to the maximum length set by Twitter, currently 140 characters.  There is currently no error checking and with text substitution some tweets may end up above the limit due to long street addresses and/or long move types.  It is not obvious that these tweets are being dropped unless you look through the logs for this type of error:

2017-02-21 13:20:07,158 [      southeast][PokeAlarm.][   ERROR] Twitter sent status 403 for URL: 1.1/statuses/update.json using parameters: (-- redacted due to privacy concerns --)
details: {u'errors': [{u'message': u'Status is over 140 characters.', u'code': 186}]}

## How Has This Been Tested?

Implemented the change and then noticed that errors stopped in my log, most tweets still worked, and that some tweets were now showing up truncated @ 140 chars, which was obvious because the google maps images don't show up if the link is corrupt

## Types of changes

## Wiki Update

Not needed

## Checklist:

### DIFF
--- a/PokeAlarm/Twitter/TwitterAlarm.py
+++ b/PokeAlarm/Twitter/TwitterAlarm.py
@@ -7,6 +7,8 @@ from twitter import Twitter, OAuth
 from ..Alarm import Alarm
 from ..Utils import parse_boolean, get_time_as_str
 
+MAX_TWEET_LEN = 140
+
 log = logging.getLogger(__name__)
 try_sending = Alarm.try_sending
 replace = Alarm.replace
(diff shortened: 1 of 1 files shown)


tomballgithub:
Wait there is a bug.  The gmaps link can make the twitter string > 140 characters.  Truncating to 140 truncates the gmaps string.  Working it...

kvangent:
Ideally we would find a solution that allows the user to choose between truncating and splitting into multiple messages. If you find a better solution feel free to open a new PR.

Issue #331, opened 2017-02-24
[BUG] Calculation of cardinal direction is not accurate


dripdropper:
## Problem Description
PokeAlarm is using an inaccurate algorithm to estimate the cardinal direction.  This can cause, for example, PokeAlarm to say a Pokemon is SE when it is clearly just S.

The solution is to use the power of mathemagics (trigonometry!) to calculate the exact bearing angle and then assign a cardinal direction accordingly.  Fortunately, I have done the hard work and developed just such an algorithm.  It's not in Python (because I currently have no good development environment), but it should be trivial to convert it and integrate it with PokeAlarm.  I have done some basic testing, and it appears to work correctly.
```
string computeDirection(degrees latitudeUser, degrees longitudeUser, degrees latitudePokemon, degrees longitudePokemon)
    radians lat1 = latitudeUser * (PI/180.0);
    radians lat2 = latitudePokemon * (PI/180.0);
    radians deltaLong = (longitudePokemon-longitudeUser) * (PI/180.0);

    double y = sin(deltaLong) * cos(lat2);
    double x = cos(lat1)*sin(lat2) - sin(lat1)*cos(lat2)*cos(deltaLong);
    degrees bearing = atan2(y, x) * (180.0/PI);
    while (bearing < 0)
        bearing = bearing + 360;

    string[8] directions = ["N", "NE", "E", "SE", "S", "SW", "W", "NW"];

    degrees adjustedBearing = (bearing + 22.5) % 360;
    int index = floor(adjustedBearing / 45.0);
    string directionToPokemon = directions[index];

    return directionToPokemon;
```

## Your Environment
Not relevant.

## Console Output
Not relevant.

## Your Alarms Config
Using `<dir>` text substitution.

## Troubleshooting Checklist
[X] Most recent version of PokemonGo-Map
[X] Most recent version of PokeAlarm
[X] Checked the Wiki page for service in question

kvangent:
This was fixed with commit 39261ad7bab10aa7b02704c51633b34b9b6d01de.

PR #332, opened 2017-02-24
Enable pure cardinal directions (N, S, E, W) in addition to NE/NW/SE/SW.

snooter:
## Description
The DIR expansion currently only shows NE/NW/SE/SW, and never pure cardinal directions (N, S, E, W). This enables all 8 possible directions.

## How Has This Been Tested?
Tested in real-life Twitter feed.

## Types of changes

## Checklist:

### DIFF
--- a/PokeAlarm/Utils.py
+++ b/PokeAlarm/Utils.py
@@ -3,7 +3,7 @@ from datetime import datetime, timedelta
 from glob import glob
 import json
 import logging
-from math import radians, sin, cos, atan2, sqrt
+from math import radians, sin, cos, atan2, sqrt, degrees
 import os
 import sys
 # 3rd Party Imports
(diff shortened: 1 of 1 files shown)


kvangent:
What is the intent of the extra 90 degrees added in?

snooter:
It's part of the algorithm - when you add that, then floor the division by 45, it dumps it in the correct bucket. There might be a way to do it without that, but I know this one works.

dripdropper:
I can confirm that this algorithm appears to be functionally equivalent to the code that I posted in issue #331.  Since mine was independently written, the algorithm is almost surely correct.

This code could be clearer if it were written in two parts.  The first part would calculate the bearing angle, and the second part would translate the angle into a direction string ("N", "NE", etc.).  The extra 90 degrees and round() stuff are used for the second part to map the angle to the directions array.

kvangent:
I guess I would prefer if we can at least source the algorithm we are using since it is fairly complex. From my understanding it should give the degrees from north, so the quarter turn is not really making any sense to me. 

I'll tested it out today to see if I'm wrong.

dripdropper:
Looking more closely at the code, this algorithm appears to have the y and x parameters of atan2() swapped, which would explain why it doesn't start from north and why there are extra manipulations needed to get the correct direction.

Are you looking for an open source library with the algorithm, or just some references to show it is correct?

kvangent:
Preferably some references to ensure it is correct.  I have a feeling it's not going to work the way I think it is going to work, but I haven't gotten to testing it yet.

dripdropper:
Here are the references I based my algorithm on:
http://stackoverflow.com/questions/3209899/determine-compass-direction-from-one-lat-lon-to-the-other
https://www.dougv.com/2009/07/calculating-the-bearing-and-compass-rose-direction-between-two-latitude-longitude-coordinates-in-php/
http://stackoverflow.com/questions/1971585/mapping-math-and-javascript
http://www.martinhansen.no/posts/20110414-getting-the-bearingheading-betweet-2-coordinates

I verified that snooter's algorithm had the same results for 100 different points I generated.

kvangent:
It seems to be correct compared to traditional equations and my own testing, I'm just concerned that I don't understand it enough and if there is a problem with in 3 months from now with it I'll be unable to diagnose it.

Give me a another day or two to see if I can find some time to sit and break it down  a bit more. It's nothing against it or you, I just want to make sure I have a little more understanding of it before I assume responsibility of it. Sorry for the wait.

PR #333, opened 2017-02-25
Split tweets into two if they exceed Twitter limit of 140 characters

tomballgithub:
## Description

Twitter currently allows tweets to be a maximum of 140 characters. There is no error checking in PokeAlarm at this time. After text substitutions, some tweets may be too long and this is not easily noticed unless logs are monitored.  I know multiple very knowledgeable people who just recently discovered missing tweets due to this issue.  This PR avoids missing tweets.

Twitter treats URLs of any size as 23 characters against the 140 limit.  So the \<gmaps\> tag translates to 23 characters and there needs to be a space in front of the URL so effectively it is 24 characters.  This leaves 116 characters for the rest of the tweet.

This code checks to see if a tweet > 140 characters with no \<gmaps\> substitution in the status string or > 116 characters if there is a \<gmaps\> substitution.  If above the appropriate limit, it will split the tweet into two and append [1/2] and [2/2] to the front of these tweets.  And because it looks better in the stream, the code puts the \<gmaps\> tag on both tweets if there was a \<gmaps\> substitution requested.

It splits the tweet by simply splitting the status string into two at the appropriate limit (140 or 116).  If the \<gmaps\> substitution was requested, that is appended to the end of the string also.  The split is not elegant but this situation should be considered an error condition since currently the tweets are simply erroring out and being lost.

## How Has This Been Tested?

Created an alerts file with 5 twitter alarms each with a status string of 200 characters and a different method of specifying \<gmaps\> to ensure all cases were covered
1.  '\<gmaps\> ' was at start of string
2.  ' \<gmaps\>' was at end of string
3.  ' \<gmaps\> ' was inside string
4.  '\<gmaps\>' was inside string (invalid status string since URL needs spaces around it)
5.  String with '\<gmaps\>' along with text substitutions

In all cases, the string was split in exactly the correct location, two tweets were generated, and the map was applied to each

I also checked strings without a \<gmaps\> substitution:
1.  200 string character with no substitutions

In all cases, the string was split in exactly the correct location, two tweets were generated, no maps were appended.

image

image

image

## Types of changes

## Wiki Update

Not needed

## Checklist:

- [X ] My code follows the code style of this project.

### DIFF
--- a/PokeAlarm/Twitter/TwitterAlarm.py
+++ b/PokeAlarm/Twitter/TwitterAlarm.py
@@ -88,6 +88,7 @@ class  TwitterAlarm(Alarm):
         else:                                                       # No gmaps link found
             maps_link_included = False
             max_length = MAX_TWEET_LEN
+            alert_str = alert['status']
             
         if len(alert_str) > max_length:
             alert_str2 = "[1/2] " + alert_str[0:(max_length-6)]     # subtract 6 due to the '[1/2] ' string length
--- a/PokeAlarm/Twitter/TwitterAlarm.py
+++ b/PokeAlarm/Twitter/TwitterAlarm.py
@@ -8,6 +8,7 @@ from ..Alarm import Alarm
 from ..Utils import parse_boolean, get_time_as_str
 
 MAX_TWEET_LEN = 140
+TWEET_URL_LEN = 23 + 1 # Twitter limit is 23 but there needs to be a preceding space before the <gmaps> tag for it to work
 
 log = logging.getLogger(__name__)
 try_sending = Alarm.try_sending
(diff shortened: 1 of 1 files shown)


PokeAlarm/Twitter/TwitterAlarm.py
```diff
@@ -70,8 +73,38 @@ def set_alert(self, settings, default):
 
     # Post Pokemon Status
     def send_alert(self, alert, info):
-        args = {"status": replace(alert['status'], info)}
-        try_sending(log, self.connect, "Twitter", self.__client.statuses.update, args)
+        # Optimize remaining tweet length when removing <gmaps>
+        # There must be spaces around <gmaps> for Twitter to recognize link properly
+        maps_link_included = True
+        max_length = MAX_TWEET_LEN - TWEET_URL_LEN
+        if ' <gmaps> ' in alert['status']:
```
kvangent: I'm not a big fan of this part. I think it would be better to just check for <gmaps> is located at the end of the string, then save it and carry it across.

PokeAlarm/Twitter/TwitterAlarm.py
```diff
@@ -70,8 +73,38 @@ def set_alert(self, settings, default):
...
+            alert_str = replace(alert['status'].replace(' <gmaps>', ''), info)
+        elif '<gmaps>' in alert['status']:                          # Even if string does not have spaces around it, go ahead and process it
+            alert_str = replace(alert['status'].replace('<gmaps>', ''), info)
+        else:                                                       # No gmaps link found
+            maps_link_included = False
+            max_length = MAX_TWEET_LEN
+            alert_str = replace(alert['status'], info)
+            
+        if len(alert_str) > max_length:
+            alert_str2 = "[1/2] " + alert_str[0:(max_length-6)]     # subtract 6 due to the '[1/2] ' string length
```
kvangent: Perhaps it would be better to make this a for loop, incase any users decide they want 3 lines worth of information.

tomballgithub:
I will work on this soon.  I will implement both changes

kvangent:
Feel free to open a new PR when ready.

PR #335, opened 2017-02-28
Youngster & Fisherman medals support

Titi9237:
## Description
Added support for Youngster & Fisherman medals notification
Added "fisherman_medal" and "youngster_medal" in the filter.json.example

## How Has This Been Tested?
Tested on my setup
The formula are based on the search made by gakushan (from TSR) : https://www.reddit.com/r/TheSilphRoad/comments/55xq49/how_size_is_determined_in_pokemon_go_plus/
I checked the formula over 10+ rattata for the Youngster medal.

## Types of changes

## Wiki Update
I add two conditions ("fisherman_medal" and "youngster_medal") in "filter.json.example" file to enable or disable each medal.

## Checklist:
- [ ] My code follows the code style of this project.


Titi9237:
It's my first work in an open-source project and I am sure I missed several codes of good practice. Feel free to make any change !

PokeAlarm/Manager.py
```diff
@@ -230,6 +230,22 @@ def handle_pokemon(self, pkmn):
...
+            return
+
+        # Check for Fisherman medal
+        if pkmn_id == 129 and filt['fisherman_medal'] is True :
+            height, weight = pkmn['height'], pkmn['weight']
+            if height/0.90 + weight/10.00 < 2.5 :
+                if config['QUIET'] is False:
+                   log.info("{} ignored: Fisherman medal condition is not fullfilled".format(name))
+            return
+
```
walaoaaa1234: you can move your code after line 263 and remove ``height, weight = pkmn['height'], pkmn['weight']``

kvangent:
Hey! I wanted to thank you for taking the time to submit this. PR's are always appreciated. 

The first issue I can see that if height and weight aren't checked to ensure they are available. If the pokemon was sent without these stats, I'm guessing it'll cause a null reference error trying to divde them by 10.

The second issue is that I don't like the idea of adding special cases for just two pokemon. I think a much more general solution would be to enabled a 'size' parameter for all pokemon ('XS', 'XL', etc), and then allow filtering based on this. It would allow any any collectors to fine tune specifications and and leave us ready for any new medals that may be added in the future. 

If you are still willing to work on this, feel free. I will leave the PR open to allow people to use it in the short term. Let me know if you have any questions about what I've stated above.

Titi9237:
@kvangent I do agree about not adding special. I'll look into it but for the moment, I have no idea how to do it.
About null reference error, I will add a condition.

kvangent:
My suggestion would be to create a new json file called "pokemon.json" and map ids to base weight and base height. With every base weight and base height in hand, calculating the size would be a simple formula.

Titi9237:
I'll give it a try !
@TheSuperBeaver It should work, if not, you can PM me

Titi9237:
So I made the "pokemon.json" file that has all the base weight and height of the Pokemon. From here I don't know how to load it in Pokealarm. @kvangent, if you have time to look at it, it would help :)

Titi9237:
I might add support for all the other medal that includes Pokemon Type !
It would simplify my view of the problem

PokeAlarm/Manager.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, name, google_key, filters, geofences, alarms, location, local
         self.__units = units
 
         # Set the minimum time_limit to send a notification (for pkmn/lures)
-        self.__time_limit = time_limit
+mod        self.__time_limit = time_limit
```
Sttug: mod ?

kvangent:
@Titi9237 I suggest you take a look a at the Utils.py section. The functions for getting move_names should be close to what you need.

Titi9237:
@kvangent I've made some change but with Breath of the Wild, I might stay away from PokemonGo for a while. I'll try to make a clean version by the end of the week

kvangent:
It looks like #338 has made good headway on this.

kvangent:
This was done in #338.

PR #338, opened 2017-03-05
Add size filter

onilton:
## Description
Add size filter for pokemon, so we can also filter to find tiny rattat or big magikarps

## How Has This Been Tested?
Tested on my own map

## Types of changes

## Checklist:

### DIFF
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -16,7 +16,7 @@ from . import config
 from WebhookStructs import Geofence
 from Utils import contains_arg, get_cardinal_dir, get_dist_as_str, get_earth_dist, get_move_damage, get_move_dps,\
     get_move_id, get_move_duration, get_move_energy, get_path, get_pkmn_id, get_team_id, get_time_as_str,\
-    parse_boolean, get_pokemon_size, get_normalized_size, Size
+    parse_boolean, get_pokemon_size, Size
 
 log = logging.getLogger('Manager')
 
@@ -521,7 +521,7 @@ class Manager(object):
                         "move_1": self.required_moves(info.get("move_1", None)),
                         "move_2": self.required_moves(info.get("move_2", None)),
                         "moveset": self.required_moveset(info.get("moveset", None)),
-                        "size": self.normalized_sizes(info.get("size", None)),
+                        "size": self.check_sizes(info.get("size", None)),
                         "ignore_missing": bool(parse_boolean(info.get('ignore_missing', ignore_missing)))
                     }
                 except Exception as e:
@@ -639,13 +639,13 @@ class Manager(object):
             list_.append(self.required_moves(moveset.split('/')))
         return list_
 
-    # Generate a list of normalized and valid sizes
-    def normalized_sizes(self, raw_sizes):
+    # Check the list of valid sizes
+    def check_sizes(self, raw_sizes):
         if raw_sizes is None:  # no sizes
             return None
         list_ = []
         for raw_size in raw_sizes:
-            size = get_normalized_size(raw_size)
+            size = raw_size
             if size in Size.available_sizes:
                 list_.append(size)
             else:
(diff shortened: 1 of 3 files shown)


PokeAlarm/Manager.py
```diff
@@ -245,7 +246,23 @@ def handle_pokemon(self, pkmn):
         height, weight, gender = pkmn['height'], pkmn['weight'], pkmn['gender']
         if gender != '?':
             gender = u'\u2642' if gender is 1 else u'\u2640' if gender is 2 else u'\u26b2' # male, female, neutral
-        
+
+        # Check the sizes of the Pokemon
+        size = None
```
kvangent: Size should really default to 'unknown' (because they may still attempt to use size as a DTS

PokeAlarm/Utils.py
```diff
@@ -148,6 +148,69 @@ def get_move_energy(move_id):
             get_move_energy.info[int(id_)] = j[id_]['energy']
     return get_move_energy.info.get(move_id, 'unkn')
 
+# Returns the base stats for a pokemon
+def get_base_stats(pokemon_id):
```
kvangent: Perhaps rather than have 3 functions, can get_base_height and get_base_width memoize themselves? I think this might make it more clear, and I know this means reading the file twice, but since it only needs to do it once I I don't think performance is an issue.

PokeAlarm/Utils.py
```diff
@@ -148,6 +148,69 @@ def get_move_energy(move_id):
...
+def size_ratio(pokemon_id, height, weight):
+    height_ratio = height / get_base_height(pokemon_id)
+    weight_ratio = weight / get_base_weight(pokemon_id)
+    return height_ratio + weight_ratio
+
+# All available pokemon sizes
+class Size:
+    XS = 'XS'
+    SMALL = 'Small'
+    NORMAL = 'normal'
```
kvangent: Can we capitalize 'normal'?

Also we should try to abbreviate all of them. Perhaps [XS, S, '', L, XL]? I would be interested to hear your thoughts.

PokeAlarm/Utils.py
```diff
@@ -148,6 +148,69 @@ def get_move_energy(move_id):
...
+    LARGE = 'Large'
+    XL = 'XL'
+    available_sizes = [XS, SMALL, NORMAL, LARGE, XL]
+
+
+# Returns normalized version of the sizes
+def get_normalized_size(raw_size):
```
kvangent: Is there a reason why we make a special class that then requires to the string equivalent? Seems like we could eliminate both this class and the method and just return the string.

onilton: Because both `get_normalized_size` and `get_pokemon_size` need that string, I did not want to have the same string repeated, risking a typo that my break stuff. Just programming best pratices. Python does not help a lot in that sense, but anyway I think it is good to have some kind of constant.

The class is to help and avoid using the `global` keyword inside the functions.

I could also move the functions into Size and make them class methods.

PokeAlarm/Utils.py
onilton: lol. That's exactly why I made the strings centralized, at least the errors would be in only one place :). I forgot to capitalize normal when moving things around.

PokeAlarm/Utils.py
kvangent: For now, I wouldn't be concerned about verify the users have put the filter in correct. I'm working on migrating the filter creation into it's own class, so I can deal with that issue then.

onilton: I don't know if I was clear, but I didn't meant user typo, I meant programmers typo (in the code itself), if someone types the string wrongly like I did with 'normal' 

If you want I can remove .`get_normalized_size` and all references from the code, but we would still need need the Size class for `available_sizes`

PokeAlarm/Manager.py
```diff
@@ -245,7 +246,23 @@ def handle_pokemon(self, pkmn):
...
+        if height != '?' or weight != '?':
```
onilton: Fixed in 	bb73be5

onilton:
@kvangent Fixed some of the issues you pointed out. Also, commented on the other points.

PokeAlarm/Utils.py
kvangent: Where else is it referenced besides inside `get_normalized_size`?

onilton: `get_pokemon_size` and https://github.com/onilton/PokeAlarm/blob/935a748bf1a47a72449e10b47cac38341712b262/PokeAlarm/Manager.py#L649

kvangent: If I am not mistaken, available_sizes is not referenced inside `get_pokemon_size`.

PR #349, opened 2017-03-16
Reusable Filters

sLoPPydrive:
## Description
Now that PokeAlarm support multiple filters the filters.json file might get a little bit crowded. Also you will usually use the same filter(s) for many different Pokemon, so if you like to modify these filter(s) you have to change them at many different places.

First of all I'd like to give things names:

**Filter** - This is usually a dictionary (or map) written with curly braces "{ ... }". All Conditions in one filter have to be true for the filter to match. This is what PokeAlarm had from the beginning. Example:
```
{"max_dist": "5000", "min_iv": "95"}
```

**Filter Set** - This is a list of multiple filters. A filter set matches if at least one of the filters of the set matches. This is the new thing for PokeAlarm 3.1 (yay!). Example:
```
[
    {"max_dist": "5000", "min_iv": "95"},
    {"max_dist": "200"}
]
```

With this PR you can give filters and filter sets names:
```
"filters": {
    "No Crap": { "min_iv": 60 },
    "Perfect": { "min_iv": 100 },
    "Nineties": { "max_dist": 5000, "min_iv": 95 },
    "Gym Gone Red": { "to_team": ["Valor"] },
    "Red Gym Lost": { "from_team": ["Valor"] },
    "Right Here": { "max_dist": "10" }
},
"filter_sets": {
    "Standard": [
        "Perfect",
        "Nineties"
    ],
    "Bubble Boy": [
        { "quick_move": ["Bubble"] },
        { "charge_move": ["Bubble Beam"] }
    ]
}
```
You can now use these names wherever a filter definition is allowed. Above you already saw that you can reference a filter by its name in a filter set. You can also...

...reference a filter or filter set by its name in a Pokemon (even the "default" one) filter configuration:
```
"pokemon":{
    "default":"No Crap",
    "Bulbasaur": "Standard",
    "Ivysaur": "Nineties",
    "Venusaur": "Bubble Boy"
}
```

...reference a filter in Gym configuration:
```
"gyms":{
    "enabled":"True",
    "ignore_neutral":"True",
    "filters": [
        "Gym Gone Red",
        "Red Gym Lost"
    ]
}
```

...and even use it in Pokestop configuration:
```
"pokestops":{
    "enabled":"False",
    "filters":"Right Here"
}
```

All this is **fully backwards compatible**. You don't need to define filters or filter sets with names, you can still use everything as before, writing down in its full length every single filter you need. And of course you can mix the traditional style with this "new" style.

In  my opinion this leads to a much cleaner filter.json file with the benefit of having your actual filter definitions centralized so if you want to modify them you do it at **one** place only.

Note that I also came across the idea of moving the filter definitions to a separate file that is being shared by all managers. For me this sounds like a good idea although I personally only use one manager. This hasn't been done yet.

## How Has This Been Tested?
My small VPS running RocketMap + PA3.

## Types of changes

## Wiki Update
Coming soon... (if there is positive feedback)

## Checklist:

### DIFF
--- a/PokeAlarm/Filters.py
+++ b/PokeAlarm/Filters.py
@@ -1,4 +1,5 @@
 # Standard Library Imports
+import copy
 import sys
 import logging
 # 3rd Party Imports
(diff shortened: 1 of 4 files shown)


kvangent:
I am on the fence about this. While I think it's a cool idea, it also adds a layer of complexity. Given that people have problems with the current format, it might introduce too many problems. 

I'm gonna think about it or a while.

sLoPPydrive:
It's all optional, so the complexity shouldn't confuse anyone until they want to. ;)

kvangent:
I've decided not to pull this in (for now). It's not that it's not a good idea - I may end up doing it later. \ I'm not really comfortable with having filters be universal between the three categories because it introduces problems with another category I'd like to finish adding (captchas). I also have some concerns about using references vs cloning for changes, (particularly in the future I'd like to have a RESTful api built again) and since that future seems way far off right now I don't really wanna force myself down any one path.

So in summary, I like the concept, but I am iffy on the implementation because it'll force me to decided a bunch of aspects I don't want to yet. (and I'm concerned about answering the billion questions because people will want to use this and some can't even figure out how JSON works at this point). 

Feel free to PM me if you have any questions or want to discuss it farther in discord.

PR #357, opened 2017-03-20
Improve Facebook integration

friscoMad:
Add options to change attachment (link) caption, description, image and name

Tested in my installation, full backwards compatible.
pokemad_2017-03-20_11-05-40

Will update Facebook wiki page after this is considered for merging 

- [x] I have updated the documentation accordingly.

### DIFF
--- a/PokeAlarm/FacebookPage/FacebookPageAlarm.py
+++ b/PokeAlarm/FacebookPage/FacebookPageAlarm.py
@@ -70,16 +70,29 @@ class FacebookPageAlarm(Alarm):
     def create_alert_settings(self, settings, default):
         alert = {
             'message': settings.pop('message', default['message']),
-            'link': settings.pop('link', default['link'])
+            'link': settings.pop('link', default['link']),
+            'caption': settings.pop('caption', None),
+            'description': settings.pop('description', None),
+            'picture': settings.pop('picture', None),
+            'name': settings.pop('name', None)
         }
         reject_leftover_parameters(settings, "'Alert level in FacebookPage alarm.")
         return alert
 
     # Post Pokemon Message
     def send_alert(self, alert, info):
+        attachment = {"link": replace(alert['link'], info)}
+        if alert['caption'] is not None:
+            attachment['caption'] = replace(alert['caption'], info)
+        if alert['description'] is not None:
+            attachment['description'] = replace(alert['description'], info)
+        if alert['picture'] is not None:
+            attachment['picture'] = replace(alert['picture'], info)
+        if alert['name'] is not None:
+            attachment['name'] = replace(alert['name'], info)
         self.post_to_wall(
             message=replace(alert['message'], info),
-            attachment={"link": replace(alert['link'], info)}
+            attachment = attachment
         )
 
     # Trigger an alert based on Pokemon info
--- a/PokeAlarm/FacebookPage/FacebookPageAlarm.py
+++ b/PokeAlarm/FacebookPage/FacebookPageAlarm.py
@@ -23,9 +23,11 @@ class FacebookPageAlarm(Alarm):
 
     _defaults = {
         'pokemon': {
-            'message': "A wild <pkmn> has appeared! Available until <24h_time> (<time_left>).",
+            'message': "A wild <pkmn> has appeared!.",
             'picture': "https://raw.githubusercontent.com/kvangent/PokeAlarm/master/icons/<pkmn_id>.png",            
-            'link': "<gmaps>"
+            'link': "<gmaps>",
+			'name': "<pkmn>", 
+			'description': "Available until <24h_time> (<time_left>)"
         },
         'pokestop': {
             'message': "Someone has placed a lure on a Pokestop! Lure will expire at <24h_time> (<time_left>).",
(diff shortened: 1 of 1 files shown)


kvangent:
I like the changes, and think they look good, but can we get the documentation updated [here](https://github.com/kvangent/PokeAlarmWiki).

Also, what are the parameters for pictures? Is it a URL to a picture? Because if so, we can add in some defaults for the rest of the fields to improve the look of the Facebook message from the get go.

friscoMad:
Sure, should I do a PR for that repo too?

Yup the picture is a URL we can use the same default as discord, I did it that way to avoid changing the current behaviour in old installations, the small map picture can be found more usefull for some people.
I think that if we provide a default there is no way to disable it in config (as None will use the default), that's the reason I thought it will be better to provide an example in docs over providing a default in code.

kvangent:
Yes, go ahead and make a PR to that wiki repo.

What we did in other alarms is setting it to 'False' turns it off.

friscoMad:
Can you point me where can I find such optional configuration?
I have little experience in python but I am unable to see anything like that in any other connector.

kvangent:
In Telegram, there are a bunch such as Location and Venue that should be about what you need.

friscoMad:
Updated with default image as I am not sure what could be the default values for other fields, and made a PR in doc repo with this new options.

kvangent:
I've got some more feedback for you (and I've poked Remedy to take a look at your wiki PR):

- Message default to `status message	A wild <pkmn> has appeared!`
- Default name to `<pkmn>`
- Description default to `Available until <24h_time> (<time_left>).`

friscoMad:
@kvangent done

kvangent:
@friscoMad You've got some tabs in there

RamiroWincor:
@kvangent My bad I did configure it only in one PC, updated without tabs.

kvangent:
Remedy and I are both just swamped with non-PA stuff right now, but I should be able to finalize this after this weekend is over.

PokeAlarm/FacebookPage/FacebookPageAlarm.py
```diff
@@ -23,15 +23,20 @@ class FacebookPageAlarm(Alarm):
...
-            'message': "A wild <pkmn> has appeared! Available until <24h_time> (<time_left>).",
-            'link': "<gmaps>"
+            'message': "A wild <pkmn> has appeared!.",
+            'picture': "https://raw.githubusercontent.com/kvangent/PokeAlarm/master/icons/<pkmn_id>.png",            
+            'link': "<gmaps>",
+            'name': "<pkmn>", 
+            'description': "Available until <24h_time> (<time_left>)"
         },
         'pokestop': {
             'message': "Someone has placed a lure on a Pokestop! Lure will expire at <24h_time> (<time_left>).",
```
kvangent: Can we get some defaults for the other alert types added? Suggestions:

pokestop:
- Message: `"Someone has placed a lure on a Pokestop!"`
- Name: `"Lured Pokestop"`
- Description: `"Lure will expire at <24h_time> (<time_left>)."`
- Caption: `None` <- Just so it is clear this is an option

Gym:
- Message: `"Team <new_team> has claimed a new gym! "`
- Name: `"It is now controlled by <new_team>."`
- Description: `"Lure will expire at <24h_time> (<time_left>)."`
- Caption: `None` <- Just so it is clear this is an option

PokeAlarm/FacebookPage/FacebookPageAlarm.py
```diff
@@ -70,16 +75,29 @@ def startup_message(self):
     def create_alert_settings(self, settings, default):
         alert = {
             'message': settings.pop('message', default['message']),
-            'link': settings.pop('link', default['link'])
+            'link': settings.pop('link', default['link']),
+            'caption': settings.pop('caption', None),
+            'description': settings.pop('description', None),
+            'picture': settings.pop('picture', default['picture']),
+            'name': settings.pop('name', None)
```
kvangent: Make sure these all take from the default correctly.

RamiroWincor:
@kvangent Done

kvangent:
Looks good, as soon as the Wiki is good I'll pull them both.

friscoMad:
I messed my identities :D 
There is one more commit that it is the picture -> image rename that I will push as soon as I test it is working.

friscoMad:
Ok now I think it is finished, I neede to change "False" to "" otherwise I would need to check against "False" or at least cast to boolean first, null in json was another option but it is not used anywhere and people already has too much troubles with json, so I think "" is the best solution.

Docs already updated also, @kvangent please tell me if you want more changes.

PokeAlarm/FacebookPage/FacebookPageAlarm.py
```diff
@@ -70,16 +82,29 @@ def startup_message(self):
...
         }
         reject_leftover_parameters(settings, "'Alert level in FacebookPage alarm.")
         return alert
 
     # Post Pokemon Message
     def send_alert(self, alert, info):
+        attachment = {"link": replace(alert['link'], info)}
+        if alert['caption']:
+            attachment['caption'] = replace(alert['caption'], info)
+        if alert['description']:
```
kvangent: I think I missed this before, but can any of these but caption actually be unset?

kvangent:
@friscoMad I agree, check from empty string seems like a good solution. Can we get an addendum somewhere in the wiki docs that makes that clear?

PR #364, opened 2017-04-02
Fix team sticker when using a language other then english.

mpw1337:
## Description

Changed how tg stickers are chosen, before they used the translated team names, which caused stickers not to get send, when using a language other then English.

## How Has This Been Tested?

Tested with various gym changes in a test telegram channel, generated by REST client.

## Types of changes

## Wiki Update

Don't think there is a need for a change in wiki, because it's just a bugfix.

## Checklist:

### DIFF
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -658,6 +658,7 @@ class Manager(object):
             "dist": get_dist_as_str(dist),
             'dir': get_cardinal_dir([lat, lng], self.__latlng),
             'new_team': cur_team,
+            'new_team_id': "Team" + str(to_team_id),
             'old_team': old_team
         })
         self.add_optional_travel_arguments(gym)
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -658,7 +658,7 @@ class Manager(object):
             "dist": get_dist_as_str(dist),
             'dir': get_cardinal_dir([lat, lng], self.__latlng),
             'new_team': cur_team,
-            'new_team_id': "Team" + str(to_team_id),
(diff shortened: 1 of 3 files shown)


PokeAlarm/Manager.py
```diff
@@ -658,6 +658,7 @@ def process_gym(self, gym):
             "dist": get_dist_as_str(dist),
             'dir': get_cardinal_dir([lat, lng], self.__latlng),
             'new_team': cur_team,
+            'new_team_id': "Team" + str(to_team_id),
```
kvangent: Can we get this changed to `"team{}".format(to_team_id)` 

(mostly to match how we do other string substitutions.

PokeAlarm/Telegram/Stickers.py
```diff
@@ -1,8 +1,8 @@
 sticker_list = {
-    'valor': u'BQADBAADGQADjGt_DbsN-MC0jtwKAg',
-    'mystic': u'BQADBAADHQADjGt_DWh6bmUsF_GeAg',
-    'instinct': u'BQADBAADGwADjGt_DcVn_7MprUVwAg',
-    'neutral': u'BQADBAADHwADjGt_DWwRwTlwYq71Ag',
+    'Team2': u'BQADBAADGQADjGt_DbsN-MC0jtwKAg',
```
kvangent: Nitpicking: I prefer lower case for this please.

PokeAlarm/Manager.py
kvangent: Can we get this changed to `"team{}".format(to_team_id)`

(mostly to match how we do other string substitutions)

Issue #201, opened 2016-09-05
[BUG] Twitter startup_message and startup_list do not display desired timezone-formatted time when user specifies timezone with -tz


reme-dy:
## Problem Description
Twitter startup message and startup list post server UTC time and ignores user-provided `-tz` argument when optional parameters `startup_message` an `startup_list` are set to `true`

## Your Environment

Operating System:  Windows 10
Python Version: (use `python -V`)  2.7.12
PokemonGo-Map startup command: `python .\runserver.py`
PokeAlarm startup command: `python .\runwebhook.py -tz America/Los_Angeles`

Did the issue persist after restarting the script?: yes

## Console Output
Timezone is set, according to log:

```
2016-09-05 11:49:10,102 [             utils] [   INFO] Timezone set to: America/Los_Angeles
```

## Your Alarms Config
`startup_message` and `startup_list` are both set to `True`.  Below is an excerpt from `alarms.json`:

```json
		{
			"active": "True",
			"type": "twitter",
			"access_token": ".",
			"access_secret": ".",
			"consumer_key": ".",
			"consumer_secret": ".",
			"startup_message":"True",
			"startup_list":"True",
			"pokemon":{
				"status": "<pkmn> in <neighborhood> at <address> <city> <postal> until <24h_time> (<time_left> remaining). <gmaps>"
			},
			"pokestop":{
				"status": "Someone has placed a lure on a Pokestop! Lure will expire at <24h_time> (<time_left>).  <gmaps>"
			},
			"gym":{
				"status":"A Team <old_team> gym has fallen! It is now controlled by <new_team>. <gmaps>"
			}
		},
```

## Troubleshooting Checklist
[x] Most recent version of PokemonGo-Map
[x] Most recent version of PokeAlarm
[x] Checked the Wiki page for service in question

kvangent:
Done with commit b2eaa649585f0d27b9abb91f4f8c008a5518fd89.

PR #369, opened 2017-04-13
Append form to name for Unown (#201) compliments RocketMap PR 1914

ifindpokemons:
e.g. Unown - B

## Description
This uses the 'form' attribute in the webhook supplied by RocketMap PR 1914 to display name - form e.g. Unown - B

## How Has This Been Tested?
My scan servers using RocketMap with 1914 applied. May not work with RocketMap not patched with PR 1914 someone should test that.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.


343334:
do you have a screenshot of it working ?

ifindpokemons:
@343334 working on that, I emulated an Unown but of course that's not as good as receiving from RM but Unown is sparse here ATM so I've put it out for all to test. Also I have a sneaking suspicion it wont work with RM that isn't patched with PR 1914. As soon as I get a confirmed Unown through the scanner and it feeds in I'll supply a screeny.

ifindpokemons:
@343334 Ah yep that KeyError is what I would expect from RM without PR 1914 as 'form' is not in the JSON object.

343334:
yea what happens is if you supress the logs it starts up normally and the servers can see it and connect fine to it as they stop reporting errors, but it doesnt actually send any notifications out, once i started an individual alarm to see how it responded then i noticed that error pop up so it just constantly keeps reporting that error pretty much crashing the notifications without you knowing it ;)  ive been meaning to do 1914 soon enough was more so waiting for this end to be completed - now I have a reason to, will get back in a day or two.

ifindpokemons:
image

Screen shot above of it working with a dummy JSON object sent to the PA endpoint, object below:

{"encounter_id": "MTA1NTUyNjUxNDM3MjAzODM1Nag=","spawnpoint_id": "6b164d1c781","form": "3","pokemon_id": 201,"latitude": -35.3047677810557,"longitude": 149.135077830091,"disappear_time": 1492153852,"last_modified_time": 1492150401,"time_until_hidden_ms": 5000,"individual_attack": 15,"individual_defense": 15,"individual_stamina": 15,"move_1": 214,"move_2": 90,"height": 1.0,"weight": 1.0,"gender": 1}}

ifindpokemons:
@343334 screen shot and sample JSON object for you to test with

ifindpokemons:
image
Confirmed with wild unown picked up by scans

kvangent:
Hey just a heads up, I'm not ignoring this, I'm just pretty busy at the moment. Here are a few notes for you before I give a more detailed review:

1. Changing the icons is outside the scope of this PR. If you want to do a different PR that's fine, but lets not mix unrelated features. 

2. Form should be not just for unknowns, but all Pokemon. This way it'll easily adapt to future Pokemon that have forms. I think that definin them in a json file is best because it allows forms to be in different languages.

ifindpokemons:
Heya yep no worries, I wouldn't recommend this merged until PR 1914 is in RM anyways.

Yep I was thinking that about the icons actually, and I'm also seeing funny image caching happening with Discord so if I resize it is still using old cache so I think I'll be best to make a new branch for images or something like that and make sure I size appropriately this time (Really new to the git stuff so learning as I go haha).

Yea I was wondering if other languages called the forms differently etc. but didn't research it, also didn't look if other Pokémon have forms. This is a bit of a durty fix definitely just to display the unown name and yep for sure it could be done way better including filtering by form. I'm working on images for now but when I get more time I'll look at revamping form to account for other Pokémon and locales. Cheers for the feedback :)

imp3tus:
Unown Form PR1914 in RM was just merged.  I've been using PR369 for about a month and it is pretty exciting.  My only recommendation is to remove the auto insert of "Unown - <form>" and leave it to the user to assign DTS <form> in their alarms before this gets merged to the main branch.  That allows the user to decide where <form> should appear.

ifindpokemons:
@imp3tus but if user puts <form> in alarms it will show unkn for pokemon that dont have form?

bbdoc:
@ifindpokemons Correct. I think more globally we need a way to hide part of the alarm text if a value is not defined. Typically something like `<pkmn [- <form>]` meaning everything within [] is only sent if <form> has an actual value. This is something we will need as well for IV and CP as far as we won't have IV and CP for all pokemons anymore.

tallypokemap:
Another way it could be done is to have form replace gender.  So if you set <gender>, it will show gender unless it's an unown, then it would show form

PokeAlarm/WebhookStructs.py
```diff
@@ -71,7 +75,8 @@ def pokemon(data):
             'gender': get_pokemon_gender(check_for_none(int, data.get('gender'), '?')),
```
tallypokemap: You could set gender to:
`'gender': get_pokemon_gender(check_for_none(int, data.get('gender'), get_form_name(int(form_raw)))),`
And remove your changes in manager.py to have form replace gender.

Arengo:
Unowns are not merged to RocketMap and this PR should be addressed to support it in PA, too.

tallypokemap:
Unowns got merged into RM the other day.... @Arengo

PokeAlarm/WebhookStructs.py
ifindpokemons: I don't like the idea of swapping one attribute for another like this, because if I am expecting a neutral gender yet suddenly I receive form instead!

ifindpokemons:
@Arengo yeh I'll try tidy it up with @kvangent's recommendations on the weekend. i.e. json file for the forms per locale and Ima look to how we can dynamically display only when data is present when specifying <form> in alarms too, seems extremely relevant now we have the case where some CP may come and others no CP.

tallypokemap:
Another option is to just have a separate manager for unowns and have a diffferent alarms.json file for unowns.  It's how I solved only listing size for rats and karps.

kvangent:
I would be okay with form returning an empty string if no form is set. (If at all possible, it would be nice to format it with '?' when still unknown but I don't know how rocketmap sends it)

tallypokemap:
The problem with returning '?' when unown is unknown is that is one of unowns actual forms @kvangent

ifindpokemons:
Ping @kvangent hey mate I've put forms inside forms.json for each locale. I'm assuming all english for now.

I have removed the dependency on #201 for form, now the user needs to specify <form> inside their alarms.json and it will show regardless of pokemon. So the trade off is users will want a seperate manager for pokemon with form to avoid "Form: unset" for example.

I'm using your function 'check_for_none' now instead of my own check so much cleaner.

Icon changes are on a seperate branch as discussed. I think I have addressed everything :D

**@everyone** needs to be aware that by default this will no longer add "- form" to the name, you will need to specify < from > for dts just as you would with anyother dts attribute.

kvangent:
@ifindpokemons Can we get this switched to v3.2 as well? Should be able to look at it within the next few days.

ifindpokemons:
@kvangent yep I'll see if I can work out how to do that in git haha.

ifindpokemons:
@tallypokemap I believe the protobuf actually defines 'unset' in the data from the encounter, I think unkn has changed to ? in this project now by the looks of it, which as you say could be confused as a legit form of #201 Unown. I think unset is good as it leave no room for confusion.

kvangent:
Moved to #376.

PR #383, opened 2017-05-05
Size tweak

tallypokemap:
## Description
This makes it so it only big karps or tiny rats get size reported.  If it is anything but a big karp or a tiny rat, it returns an empty string ''.  This makes it so you don't need a separate manager to report size.

## How Has This Been Tested?
I've tested on my setup and works well

## Types of changes

## Checklist:

### DIFF
--- a/PokeAlarm/Utils.py
+++ b/PokeAlarm/Utils.py
@@ -207,6 +207,8 @@ def get_pokemon_size(pokemon_id, height, weight):
         return 'tiny'
     elif size > 2.5 and pokemon_id == 129:
         return 'big'
+    else:
+	return ''
 
 
 # Returns the gender symbol of a pokemon:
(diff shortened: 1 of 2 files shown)


kvangent:
So this isn't going to happen because it reduces the functionality. Right now it may be a little more of a pain to use a separate manager for Rattata/Magikarp to have the size listed, but all other Pokemon can have their size reported. This may be useful for future medals or even just collected that want all the different sizes.

tallypokemap:
@kvangent It would be odd to try to collect sizes since there is no way in-game to see any of their sizes without appraisal or calculation, but if you feel this still reduces fuctionality, would you be opposed to me adding in a config command 
```
rat_karp_size_only
```
that will be false by default but if set to true, then this feature is enabled?

PR #389, opened 2017-05-09
Updated magikarp/gyrados, 2nd gen pokemons & gyms icons

MARTINI1:
## Description

I long ago noticed that the icons of the gyms were horrible and now with the beta of v3.2 I have decided to renew the icons, apart from changing the images of the 2nd gen to sprites of Pokemon Go as the 1st gen.

## How Has This Been Tested?

In my own pokealarm with branch 3.2.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #402, opened 2017-07-05
Raids

fosJoddie:
## Description
Adds support for sending out alarms for upcoming raids (eggs) and active raids (raid pokemon).

Added two new filter sections: for eggs and raids :
- Filter by raid level for eggs
- Filter by individual pokemon for raids

Added two new alarms:
- Egg: Sends alert about an upcoming raid - includes raid start time, end time and level
- Raid: Sends alert about a raid that has started - includes pokemon, raid level, end time and cp

Refactored filter check for pokemon to reuse the same filter check for both raids and pokemon alarms.

Added a None check in string replace that should handle some weird errors in very specific revese geocode address lookups.

Fixed a spelling mistake in the section for loading pokemon filters where it checked against the wrong collection (I think)

## How Has This Been Tested?
Tested against my personal Monocle fork and the current raid PR from RocketMap with alarms sending to Discord.
Tested that refactoring of pokemon filtering did not break pokemon alerts
Tested that both egg and raid alerts work and does not get ignored or failed
Various user reports confirm that other alarm targets also work (like twitter), but they are untested by me personally.

## Types of changes

## Wiki Update
Excerp below, see also : <https://github.com/fosJoddie/PokeAlarm/wiki/Getting-started-with-raid-alarms>
To use raid alarms, existing filters.json file must be updated with the new section describing raids in filters.json.example.

The filter following filters are available:

- **min_level**:  An integer describing the minimum level of the raid you want to send alarms about. Raids with a lower level will not be sent
- **max_level**: An integer describing he maximum level of the raid you want to send alarms about. Raids with a higher level will not be sent
- **ignore_eggs**: A boolean describing whether or not alarms should be sent if the raid does not yet have an active pokemon battle.

The default filter will for the most part be ignored, but is kept the same way as pokemon filters for simplicity. Raid pokemon will only report CP and have perfect IVs.  See the pokemon filter documentation for more info about the other fields.

if you want custom alarm messages you can use text replacing as with existing alarms. There are two new messages to set up:
- **raid** - Alarm when a pokemon battle has started in a gym. Contains raid level and pokemon information
- **egg** - Alarm when a raid battle is announced - only contains raid level and no pokemon

For raid messages you can use the same keywords as pokemon alarms, including cp and moves.

Additionally there are a few new keywords:
- <raid_level>:  Puts in the level of the raid
- <begin_24h_time>, <begin_12h_time>, <begin_time_left> the time for the raid start in 24 hour, 12 hour and time left until start.

## Checklist:


CatchMeImFamous:
working very good for me , except <raid_level> returns a "?"

fosJoddie:
@CatchMeImFamous - it will return a ? if the implementation that sent the webhook did not contain the raid level information.  That's not a PA error

jokid64:
Ow, I forked your ...

nhumber:
works great.

rubendgpedrosa:
Working like a charm.

PokeAlarm/Manager.py
```diff
@@ -740,6 +755,113 @@ def process_gym(self, gym):
...
+            filters = self.__raid_settings['filters'][pkmn_id]
+            passed = self.check_pokemon_filter(filters, atk, def_, sta, quick_id, charge_id, cp, dist, None, None, iv, level,
+                                               name, None)
+            # If we didn't pass any filters
+            if not passed:
+                log.debug("Raid {} did not pass pokemon check".format(id_))
+                return
+
+        # check if the level is in the filter range or if we are ignoring eggs
+        passed = self.check_raid_filter(self.__raid_settings,raid)
```
neskk: I would probably check the raid filters before the pkmn ones because they're quicker and will often prevent having to check pkmn filters.

fosJoddie: agreed, moved to above pokemon filter check

fosJoddie:
@kvangent made a wiki PR for the raid changes, hope that helps

tallypokemap:
Unless I'm missing something obvious, this PR doesn't support geofence DTS for raids.  It might be good to add that, but it might just as easily be added in a separate PR

PokeAlarm/Manager.py
```diff
@@ -740,6 +775,113 @@ def process_gym(self, gym):
...
+        if len(self.__geofences) > 0:
+            inside = False
+            for gf in self.__geofences:
+                inside |= gf.contains(lat, lng)
+            if inside is False:
+                if self.__quiet is False:
+                    log.info("Raid update ignored: located outside geofences.")
+                return
+        else:
+            log.debug("Raid inside geofences was not checked because no geofences were set.")
```
hokiepokedad: Consider being consistent with the other process_* functions by using the self.check_geofences(name, lat, lng) function.  This will also allow for the geofence name to be process as a DTS token.

hokiepokedad: `
raid['geofence'] = self.check_geofences('Raid', lat, lng)
if len(self.__geofences) > 0 and gym['geofence'] == 'unknown':
    log.info("Raid rejected: not inside geofence(s)")
    return
`

fosJoddie:
@tallypokemap  @hokiepokedad  

Updated so geofences are checked like in other places and will include the name of first matching one for DTS purposes now.

Ben-13:
Is gyms enabled: true required to get raid boss info? I've got everything setup and seemingly working but i am only getting egg info on discord, never what hatches.

DragonHeart69:
@Ben-13 You need the code from filters.json.example from line 291 to the last. copy this to you're filters.json & i'ts work for me.
https://github.com/kvangent/PokeAlarm/pull/402/files#diff-7edc9ce7d6fd3caf1ef9f0ad80b41372

Ben-13:
that did it @DragonHeart69 , thank you. I misunderstood its need.

fosJoddie:
Good you solved it, but please don’t comment on the PR with support questions

Ben-13:
Alright. Perhaps new to GitHub as it says discussion I thought asking to be
ok.

Qualimiox:
While this is a great PR overall and it's working well on one of my setups, I was having issues on the other one and didn't have any means of debugging because the only logs for raids right now are the "Setting up Raid filters" one....I'd like to see specific logs for each incoming raid from the webhook, like we currently have for pokemon.

fosJoddie:
@Qualimiox by setting the debug flag you would have gotten those messages. 

But have changed the log level for the raid messages to be in line with how the other messages are doing it. You should see processing info about raid as well now, with info on what made it not trigger (like wrong level etc)

MARTINI1:
@fosJoddie you thought how to unify the configurations of images ?

As the "icon_url" in the slack and discord or the "image" in facebook that in the raid section points to the master branch of kvangent and in the section egg points to your own repo

Another thing to think is that if this pr is merged will not use the the images if they are as master until you pass the branch v3.2 to master

I say it because in my PR "updated icons" #389 are unused while the branch 3.2 does not merge with master and I see this PR will happen something similar

To find for a fix to this I think that would have to make a temporary patch to load the images of the same branch v3.2 and then when it is already passed to master make a final PR fixing everything and leaving everything definitive

jophj:
Should this work with the current version of RocketMap? I see notifications for unhatched raids, but I can't see notifications for hatched ones. I also get some  `[    MainProcess][WebhookStr][   ERROR] Invalid type specified (raid). Are you using the correct map type?`

As I said, I'm with current RocketMap version.

dongemus:
@jophj It works fine, you did something wrong. This isn't an area for support, there is no support for PRs . Unfortunately, you'll have to figure it out on your own or wait for it to be merged with a supported branch.

filters.json.example
```diff
@@ -288,5 +288,270 @@
         "Lugia":"False",
         "Ho-Oh":"False",
         "Celebi":"False"
+    },
+    "raids":{
+        "enabled":"True",
+        "filters": {
```
kvangent: Can we name this to 'eggs' instead of filters?  Filters doesn't seem quite appropriate as the terminology.

PokeAlarm/Manager.py
```diff
@@ -399,11 +370,11 @@ def process_pokemon(self, pkmn):
                 log.debug("Pokemon IV percent was not checked because it was missing.")
 
             # Check the Attack IV of the Pokemon
-            if atk != '?':
-                if not filt.check_atk(atk):
+            if attack is not None and attack != 'unkn' and attack != '?':
```
kvangent: Why rename the variables?

PokeAlarm/Manager.py
```diff
@@ -298,54 +304,19 @@ def clean_hist(self):
...
+        # raid history has a different structure because it saves both expire time and pokemon
+        old = []
+        for id_ in self.__raid_hist:
+            if self.__raid_hist[id_]['expire_time'] < datetime.utcnow():
+                old.append(id_)
+        for id_ in old:     # Remove expired raids
+            del self.__raid_hist[id_]
+
+    # Check if a given pokemon is active on a filter
+    def check_pokemon_filter(self, filters, attack, defense, stamina, quick_id, charge_id, cp, dist, form_id, gender, iv,
```
kvangent: I don't think it makes sense to expand the variables out for this.

PokeAlarm/Manager.py
```diff
@@ -360,7 +331,7 @@ def process_pokemon(self, pkmn):
                 log.debug("Filter dist was not checked because the manager has no location set.")
 
             # Check the CP of the Pokemon
-            if cp != '?':
+            if cp is not None and cp != '?' and cp != 'unkn':
```
kvangent: I don't think it is appropriate to check if variables == None here. I touch base on this more in webhook structs.

PokeAlarm/WebhookStructs.py
```diff
@@ -114,6 +116,60 @@ def pokestop(data):
...
+    def raid(data):
+        log.debug("Converting to raid: \n {}".format(data))
+
+        quick_id = check_for_none(int, data.get('move_1'), '?')
+        charge_id = check_for_none(int, data.get('move_2'), '?')
+
+        raid_end = None
+        raid_begin = None
+
+        if 'raid_begin' in data:
```
kvangent: Why are we treating raid and eggs as the same? While they have some similarities, Raids are essentially Pokemon with extra steps while eggs have far less variables. 

I think a much cleaner way to handle this would be to treat eggs and raids as different types of alerts. We need to detect which it is and construct a different struct for each. Each should be handled differently throughout the rest of the process. You did it this way for the alarms (raid vs raid_alert), but I think it needs to go a step farther and start here.

filters.json.example
```diff
@@ -288,5 +288,270 @@
...
+            "max_level": "5",
+            "ignore_eggs": "False"
+        },
+        "default": {
+            "min_dist":"0", "max_dist":"inf", "min_cp": "0", "max_cp": "999999", "min_iv":"0", "max_iv":"100",
+            "min_atk": "0", "max_atk":"15", "min_def": "0", "max_def":"15", "min_sta": "0", "max_sta":"15",
+            "quick_move": null, "charge_move": null, "moveset": null,
+            "size": null, "gender": null, "form": null, "ignore_missing": "False"
+        },
+        "Bulbasaur": "True",
```
kvangent: We should get rid of the Pokemon that aren't currently used in raids. When/if Pokemon get rotated we can update the example or people can change their configs accordingly.

fosJoddie:
I refactored so the egg and raid is separated at an earlier stage.

There's now two issues with the PR

- both raids and raid alert (egg) share the same filter - I'm torn between adding another filter for the eggs alone or not. If there's another filter, it's possible to filter out level 2 eggs but still get warnings about some level 2 raids once they spawn. 

- the URL changes could be done as a last push before merge or simply as a separate PR after this one.

kvangent:
I would change it to a separate filter, and simple do a history check for eggs in the raid handler.

I would just ignore the url changes for now - probably best to wait until post merge.

filters.json.example
```diff
@@ -288,5 +288,42 @@
...
+        "Jolteon":"True",
+        "Machamp":"True",
+        "Vaporeon":"True",
+        "Blastoise":"True",
+        "Charizard":"True",
+        "Lapras":"True",
+        "Rhydon":"True",
+        "Snorlax":"True",
+        "Tyranitar":"True",
+        "Venasaur":"True"
```
MARTINI1: error in line 327:

> [    MainProcess][   Filters][   ERROR] Unable to find pokemon named 'Venasaur'...
[    MainProcess][   Filters][   ERROR] Please see PokeAlarm documentation for proper Filter file formatting.

replace to correct name Venusaur

fosJoddie:
I have moved egg filters out in their own section of Filter. It is now simply min and max raid level.

Likewise the raid section is now only filtered by specific pokemon and as such is much "easier" to understand

PokeAlarm/Manager.py
```diff
@@ -373,7 +364,7 @@ def process_pokemon(self, pkmn):
                 log.debug("Pokemon 'cp' was not checked because it was missing.")
 
             # Check the Level of the Pokemon
-            if level != '?':
+            if level is not None and level != 'unkn' and level != '?':
```
kvangent: Why do we want to move to checking for these three things? level should never be None at this point

PokeAlarm/Manager.py
```diff
@@ -740,6 +796,185 @@ def process_gym(self, gym):
...
+            old_raid_end = self.__raid_hist[id_]['raid_end']
+            if old_raid_end == raid_end:
+                if self.__quiet is False:
+                    log.info("Raid {} ignored. Was previously processed.".format(id_))
+                return
+
+        self.__raid_hist[id_] = dict(raid_end=raid_end, pkmn_id=0)
+
+        # don't alert about expired raids
+        if datetime.utcnow() > raid_end:
```
kvangent: This should really use time_left, similar to pokemon to allow users to not receive notifications if there isn't enough time left remaining.

PokeAlarm/Manager.py
```diff
@@ -740,6 +796,185 @@ def process_gym(self, gym):
...
+
+        pkmn_id = raid['pkmn_id']
+        raid_end = raid['raid_end']
+
+        # raid history will contain the end date and also the pokemon if it has hatched
+        if id_ in self.__raid_hist:
+            old_raid_end = self.__raid_hist[id_]['raid_end']
+            old_raid_pkmn = self.__raid_hist[id_].get('pkmn_id', 0)
+            if old_raid_end == raid_end:
+                if old_raid_pkmn == pkmn_id:  # raid with same end time exists and it has same pokemon id, skip it
```
kvangent: There can only be one raid per gym at a single time right? Why are we checking pkmn_id?

PokeAlarm/Manager.py
fosJoddie: I was running some code that had cp and level sneak through as None and unkn - but after fixing that problem I forgot to change these monster if statements ... 🤦‍♀️

PokeAlarm/Manager.py
fosJoddie: the cache is updated from both eggs and raids since they share the source (a gym raid) the history can either contain only a raid end or raid end + pokemon. 

Seemed like the better option to me to just keep one history lookup instead of splitting something that must be checked in both places anyway.

PokeAlarm/Manager.py
fosJoddie: had it as a todo in my mind, good catch.

kvangent:
I'm going to pull, but please try to fix your mistakes in the wiki as soon as possible.

PR #407, opened 2017-07-18
Add support for gym DTS into egg and raid alarms

fosJoddie:
## Description

WIP 
Adds support for gym details in DTS for raid and eggs

* `<gym_name>`
* `<gym_description>`
* `<gym_url>`

Requires RM to scan gym details and send it

## How Has This Been Tested?
Tested locally

## Types of changes

## Wiki Update
Will update DTS page with gym info for eggs and raids

## Checklist:

### DIFF
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -1010,20 +1010,13 @@ class Manager(object):
 
     # Add gym details to an info object
     def add_gym_details(self, info, gym_id):
-        if gym_id in self.__gym_info:
-            gym_info = self.__gym_info[gym_id]
-            info.update({
-                "gym_name": gym_info.get('name', 'unknown'),
-                "gym_description": gym_info.get('description', 'unknown'),
-                "gym_url": gym_info.get('url', 'https://raw.githubusercontent.com/kvangent/PokeAlarm/master/icons'
-                                               '/gym_0.png')
-            })
-        else:
-            info.update({
-                "gym_name": 'unknown',
-                "gym_description": 'unknown',
-                "gym_url": 'https://raw.githubusercontent.com/kvangent/PokeAlarm/master/icons/gym_0.png'
-            })
+        gym_info = self.__gym_info.get(gym_id,{})
+        info.update({
+            "gym_name": gym_info.get('name', 'unknown'),
+            "gym_description": gym_info.get('description', 'unknown'),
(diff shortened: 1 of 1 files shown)


PokeAlarm/Manager.py
```diff
@@ -679,23 +679,26 @@ def process_pokestop(self, stop):
             thread.join()
 
     def process_gym(self, gym):
+        gym_id = gym['id']
```
sLoPPydrive: I would put the `if self.__gym_settings['enabled'] is False:` check at the very beginning and remove the `if self.__gym_settings['enabled'] is True:` later on.

fosJoddie:
@sLoPPydrive is Github being coy or did you remove your comment? 

In any case - because of raid and egg alarms needing cached gym info, we check for either of these to be true to make sure info is saved. Then later return if we dont alert gyms.

sLoPPydrive:
Yes, I removed the comment because I noticed my mistake. :)

PokeAlarm/Manager.py
```diff
@@ -999,6 +1006,15 @@ def add_optional_travel_arguments(self, info):
         if self.__api_req['DRIVE_DIST']:
             info.update(**self.get_driving_data(lat, lng))
 
+    # Add gym details to an info object
+    def add_gym_details(self, info, gym_id):
```
fosJoddie: yeah I'm not familiar enough with Python to think of those cases. 

But why can't that method still stay to minimize code copy paste? Just with your suggested changes

kvangent: I guess the reason I don't think it is worth moving it into a function is because right now the function takes up about 6-9 lines of code (depending on how you count) where as it would only be around 6 lines to do it without it. The additional complexity doesn't seem worth the trade-off.

PokeAlarm/Manager.py
```diff
@@ -17,11 +17,11 @@
     load_raid_section
 from Utils import get_cardinal_dir, get_dist_as_str, get_earth_dist, get_path, get_time_as_str, \
     require_and_remove_key, parse_boolean, contains_arg
+
 log = logging.getLogger('Manager')
 
 
 class Manager(object):
```
kvangent: Can we put the line back under here?

fosJoddie:
I'm not a Python developer. I have no issue inlining the method and leaving it like that.

PR #425, opened 2017-08-29
Introducing generate_geofence.py based on OSM

KartulUdus:
## Description

Intorducing "generate_geofence.py".
With this utility it's possible to generate a geofence file automatically by entering the filename + lcoation name.
For example: 
`python generate_geofence.py -dir mygeofence.txt -l berlin`
would create a geofence in the directory of "generate_geofence.py"  with the exact borders of Berlin.

Multi-polygons:
Some areas consist of multiple polygons. 
"generate_geofence.py" detects this and is able to iterate through each polygon and write the geofence with multiple areas.
By default it would pick a single (first) polygon from a multi-polygon; however, if you wish to use the multiple geofences in the same file, use the `-mp` flag.

Example:
`python generate_geofence.py -dir test.txt -l japan -mp`
would generate 24 separate areas in the same file iterating the area name "[japan <n + 1>]".

When there is no ["osm_type":"relation"] type element returned: 

For some locations, such as "Manhattan central park", there is no "relational" polygon available. 
should this happen to be the case, you can try finding a "way" type polygon by the `-w` flag.

`python generate_geofence.py -dir test.txt -l 'Manhattan central park' -w`

You can check for the existence of a polygon on 
https://nominatim.openstreetmap.org
You can check the coordinates of the generated file in:
https://www.mapcustomizer.com/ -> bulk entry -> copypasta contents of geofence file.

## Motivation and context: 

Generating configs and geofences for multiple managers in PokeAlarm is tedious.
Automation of geofence generation makes it easier for users to specify their areas.

## How Has This Been Tested?
* Generated multipolygon and single polygon geofences of various areas with relational polygons and "ways" with different amount of points
* Python 2.7.12 & pip 9.0.1
* Generated geofence files on Windows/OSX/linux
* Ran PokeAlarm on home desktop(windows) and amazon t2.micro instance on ubuntu 16.4
* POST'ed fake mewtwos sending alarms to discord. 
    Included arrows with a narrow pier, Included secondary polygons (islands)
* Tried large multipolygon areas - ``-l russia -mp` was a total of 30 polygons and 98k coords
    This was generated by the t2.micro instance in a few seconds and PA seemed to accept it without an issue.

I have *not* used such a large geofence with standard traffic (sending all wh_types; from RM for instance)

## Types of changes

## Checklist:

### DIFF
--- a/generate_geofence.py
+++ b/generate_geofence.py
@@ -1,0 +1,88 @@
+# -*- coding: UTF-8 -*-
+import json
+import urllib2
+import csv
+import configargparse
+
+
+def main():
+
+    # ARGS
+    parser = configargparse.ArgParser(description='geoFenceGen')
+    parser.add_argument(
+        '-dir',
+        '--directory',
+        help='Directory and filename of geofence. ex: geofence/Bratislava.txt',
+        required=True)
+    parser.add_argument(
+        '-l',
+        '--location',
+        help='City/Country/area you would like a geofence for',
+        required=True)
+
+    parser.add_argument(
+        '-w',
+        '--way',
+        help=('Use if there\'s no relational polygon' +
+              '(ex: Manhattan central park'),
+        action='store_true',
+        default=False)
+    parser.add_argument(
+        '-mp',
+        '--multiPolygon',
+        help=('Use if you wish to write all polygons of an area'),
+        action='store_true',
+        default=False)
+    args = parser.parse_args()
+    loc = (args.location)
+    loc = loc.replace(" ", "+")
+    end = ("https://nominatim.openstreetmap.org/"
+           "search.php?q='{}'&polygon_geojson=1&format=json".format(loc))
+
+# Check for good response and get polygon
+
+    if (len(json.load(urllib2.urlopen(end))) < 1):
+        print 'Unable to find areas, exiting'
+        exit(1)
+    else:
+        raw = json.load(urllib2.urlopen(end))
+        mode = 'way' if (args.way) else 'relation'
+        coords = [x for x in raw if x['osm_type'] ==
+                  '{}'.format(mode)][0]['geojson']['coordinates']
+
+# check if area has single polygon and write
+
+    if (len(coords[0][0]) == 2 or (args.way)):
+        # Write fenceFile
+        l = [["{}".format(args.location)]]
+        with open('{}'.format(args.directory), 'wb') as f:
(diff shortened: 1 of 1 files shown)


generate_geofence.py
```diff
@@ -0,0 +1,88 @@
...
+    parser.add_argument(
+        '-mp',
+        '--multiPolygon',
+        help=('Use if you wish to write all polygons of an area'),
+        action='store_true',
+        default=False)
+    args = parser.parse_args()
+    loc = (args.location)
+    loc = loc.replace(" ", "+")
+    end = ("https://nominatim.openstreetmap.org/"
```
fosJoddie: End of what? Variable should ve called url or osm_url or something more descriptive.

generate_geofence.py
```diff
@@ -0,0 +1,88 @@
...
+import configargparse
+
+
+def main():
+
+    # ARGS
+    parser = configargparse.ArgParser(description='geoFenceGen')
+    parser.add_argument(
+        '-dir',
+        '--directory',
```
fosJoddie: A directory + filename is called a path, or simply file. I would use file or out/outfile as arg name

generate_geofence.py
```diff
@@ -0,0 +1,88 @@
...
+import json
+import urllib2
+import csv
+import configargparse
+
+
+def main():
+
+    # ARGS
+    parser = configargparse.ArgParser(description='geoFenceGen')
```
fosJoddie: The Description should explain a bit more

kvangent:
While a nifty script, my 2 cents is that this would be beyond the project scope and would be bloat.

Also, you probably want this PR to go against the dev branch.

generate_geofence.py
KartulUdus: I like that. Can change

generate_geofence.py
KartulUdus: short for end*point*. can change if matters

KartulUdus:
Overall, I wanted to make something that would automate writing all the needed configs/filters/alarms/managers and the geofence creation seemed to be hardest as it needed 3rd party data.

Perhaps it's a bloat in the root of the project, but I believe PA could use a `tools/` folder to automate much setup in the future

KartulUdus:
It didn't pass a full local stress test with 16 managers and 8 cities. 

the output needs magic to make it lightweight and  I may have a few ideas. 
closing for now

Issue #328, opened 2017-02-22
[ENHANCEMENT] cache reverse location


evenly-epic-mule:
## Enhancement Description
add persistant cache for reverse location
mapping `{spawn_id: address data}`

## Use cases
reverce location?
save requests on the Google API

## Other Links
for saving the data we could use pickle

evenly-epic-mule:
would help for #325

pogo-excalibur:
Do be mindful of this: https://developers.google.com/maps/terms#section_10_5
(Specifically the 'No caching or storage' section).
If you read the document, caching to reduce the number of requests sent should be fine so long as the cache is only persisted for up to 30 days.

kvangent:
This was pulled in [Patch v3.3](d779668a9b63e1326d8b187f3ac806d538472fb3).

PR #437, opened 2017-09-30
Pickle cache

fosJoddie:
## Description
Introduces two new options for PA:

`use-adr-file-cache`: Cache google address lookups in a file cache. Since most maps have very static spawn point and gym position data, the need to look up addresses repeatedly is limited and it makes sense to save requests by caching the information.
`use-gym-file-cache`: Cache gym details in a file cache so that a freshly started PA do not have to rely on RM for getting gym information.

- Cached information is automatically saved every 10 seconds, or if PokeAlarm is exited with Ctrl+C or any other graceful method.

address DTS has been changed a tiny bit. It now will return an empty value if street name or street number is unknown. This allows for streets that have no numbers (like squares) to show up correctly.
In addition there's a new DTS called `<address_eu>` that will write out "Alphabet Street 123" as addresses is usually written in Europe/Most of the world(?)

Lastly this also include a minor change where the gym processing would check if it was inside geofence twice.

I realize this PR is not compatible with current dev branch, but I made this one already back before dev had caching so I found it easier to finish this one instead of trying to merge what I did with what dev has. 

## How Has This Been Tested?
Tested locally for an extensive time with no issues. 
Been running for 1 month with 10 managers with alarms for pokemon and raids. tested both with imported cache files and fresh ones.
Not tested anything of gyms, but should be unchanged

## Types of changes

## Wiki Update
Need to update the config section on how to enable the cachings and also with information about where to go to clear it if wanted.

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *.ini
 *.txt
 !locales/*.json
+/.pickles/
--- a/PokeAlarm/Cache.py
+++ b/PokeAlarm/Cache.py
@@ -1,0 +1,63 @@
+import pickle
+import os.path
+import errno
+import logging
+
+log = logging.getLogger('Cache')
+
+
+# Cache helper object
+# Caches the reverse geocode lookup of coordinates to addresses
+# and the gym details
+class Cache(object):
+    def __init__(self, name, use_adr_cache, use_gym_cache):
+        self.adr_cache = {}
+        self.gym_cache = {}
+        self.__name = name
+        self.__use_adr_cache = use_adr_cache
+        self.__use_gym_cache = use_gym_cache
+        self.__pkl_file_adr = '.pickles/cached_{}_adr.pkl'.format(name)
+        self.__pkl_file_gym = '.pickles/cached_{}_gym.pkl'.format(name)
+
+    # load the cache
+    @staticmethod
+    def load_pickle(file_name):
+        try:
+            with open(file_name, 'r+') as pickle_file:
+                cache = pickle.load(pickle_file)
+                log.info("Loaded cache with {} entries".format(len(cache)))
+
+            return cache
+        except (OSError, IOError):
+            log.warn("Unable to load the pickle {}, starting fresh".format(file_name))
+            return {}
+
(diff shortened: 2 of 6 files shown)


Goldclownfish:
Caching is against googles tos.

pogo-excalibur:
From my comment on #328:

Do be mindful of this: https://developers.google.com/maps/terms#section_10_5

(Specifically the 'No caching or storage' section).

If you read the document, caching to reduce the number of requests sent should be fine so long as the cache is only persisted for up to 30 days.

(I believe this restriction may be relaxed if you are using a premium API key).

Goldclownfish:
No caching or storage. You will not pre-fetch, cache, index, or store any Content to be used outside the Service, except that you may store limited amounts of Content solely for the purpose of improving the performance of your Maps API Implementation due to network latency (and not for the purpose of preventing Google from accurately tracking usage), and only if such storage:
is temporary (and in no event more than 30 calendar days);
is secure;
does not manipulate or aggregate any part of the Content or Service; and
does not modify attribution in any way.     

Don’t think u can get around this because caching this has nothing to do with latency issues

fosJoddie:
Meanwhile people use multiple free google keys, with no concern. 

I’m fine if it isn’t allowed, but: 

`You will not pre-fetch, cache, index, or store any Content to be used outside the Service` Says: used OUTSIDE the service. I am not concinced limited caching falls outside their terms. 

Anyhow I am going to rebase this against the dev branch when I can, which btw already caches geocode results in memory, fyi.

kvangent:
Just to weigh in here, my interpretation is that caching in memory is fine because it is temporary to a degree (it seems unlikely PA will be running for over 30 consecutive days) but writing for a file may break that use case.

Also, the reason we haven't added in rotating keys is because I want to be wary of abuse. I added in one per manager because I felt the separation of those managers is close enough to 'different projects' and not different than launching multiple instances of PA. 

I think caching the gym info to file is a great idea (one that I've tried to implement in the gym_cache branch) and would love to incorporate it into PA.

PR #447, opened 2017-10-10
Save gym details to file between sessions

fosJoddie:
## Description
Introduces a new option for PA:

`use-gym-file-cache`: Cache gym details in a file cache so that a freshly started PA do not have to rely on RM for getting gym information.

- Cached information is automatically saved every 10 seconds, or if PokeAlarm is exited with Ctrl+C or any other graceful method.

This also include a minor fix to where the gym processing would check if it was inside geofence twice.

## How Has This Been Tested?
Tested locally for an extensive time with no issues. 
Been running for 1 month with 10 managers with alarms for pokemon and raids. tested both with imported cache files and fresh ones.
Tested after rebase into dev branch with no new problems
Tested in Windows Server 2012 (!) and Windows 10.

## Types of changes

## Wiki Update
Need to update the config section on how to enable the cachings and also with information about where to go to clear it if wanted.

## Checklist:


fosJoddie:
Since rebasing this meant I had to remake the PR, I'm going to pull in the comments from the last PR about google not allowing caching or reverse geocode lookups (or any other access to their service)

Kvangent wrote:

> Just to weigh in here, my interpretation is that caching in memory is fine because it is temporary to a degree (it seems unlikely PA will be running for over 30 consecutive days) but writing for a file may break that use case.
>
>Also, the reason we haven't added in rotating keys is because I want to be wary of abuse. I added in one per manager because I felt the separation of those managers is close enough to 'different projects' and not different than launching multiple instances of PA.
>
>I think caching the gym info to file is a great idea (one that I've tried to implement in the gym_cache branch) and would love to incorporate it into PA.

The google Terms clearly states: 
> You will not pre-fetch, cache, index, or store any Content to be used outside the Service

the wording here is very clear. It does not allow you to cache the results and it does not allow you to store it. The key part of this sentence is what is meant by "to be used outside the Service". I interpret this as meaning that caching (in memory or otherwise) up to 30 days is acceptable, but you aren't allowed to cache it for use in other projects/services. Doing all addresses in New York and then publishing a website listing all those addresses and coordinates would fall into such a use case. 

I may be wrong though, but it is definitely not clear written, and that is very much on purpose.

For the rest of the PR, please do test and comment, removing geocode caching is just a matter of a few lines of code.

kvangent:
In regards to the geocoding file caching, it also says this [here](https://developers.google.com/maps/documentation/geocoding/geocoding-strategies#caching)

> The Google Maps APIs Terms of Service allow you to cache geocodes (that is, store them on your server for a limited period). Caching can be useful if you have to repeatedly look up the same address. 

So it seems clear you are allowed cache, and to do it to reduce the number of redundant API calls made. The problem I have with moving the cache to a file is that it then becomes less temporary. There is nothing stopping users from keeping their caches for more than 30 days. In fact, with this model it is the default behavior - unless the user goes in and manually clears the cache, it will stick around. 

As a side note (and more of a meta comment) is that I need to find a way to establish a policy advising best practice is to limit PRs from having a wide scope. I feel like this PR is really close to that line if not slightly over it - there are several features in here that could probably be in their own PRs (gym caching, address caching, graceful process exiting, and the changes to the address DTS). I understand that these are all tangentially related but it does complicate the process of reviewing the PR.

fosJoddie:
I removed the permanent saving of addresses and I also removed the changes to address DTS. 

The address files would need to be purged > 30 days or similar to be within googles terms, but off topic note, a long running pokealarm would/should actually purge old addresses from memory as well.

pogo-excalibur:
Or we could not use Google for geocoding?

I currently use https://geocode.xyz for a couple of projects, and it seems to be working out OK.

Terms:
```
You may store and use the data we provide in any way you see fit, as long as you don not re-sell these data as is (we do however provide reseller licenses on a case by case basis. Contact us for more information.) 
```

kvangent:
The API limit here is 1 per second, which means throttling would need to be incorporated. Have you used this service? Is it any good?

fosJoddie:
Since we're already off topic, LocationIQ is free with 60 requests per minute. https://locationiq.org/#pricing and uses OSM backed data.

PokeAlarm/Cache.py
```diff
@@ -0,0 +1,56 @@
...
+    # Load the cache from pickle files
+    def load(self):
+        if self.__use_gym_cache:
+            self.gym_cache = self.load_pickle(self.__pkl_file_gym)
+
+    # Save the cache into pickle files.
+    def save(self):
+        if self.__use_gym_cache and not os.path.exists('.pickles'):
+            try:
+                os.makedirs('.pickles')
```
kvangent: This could give permission issues. One solution is to check for the folder on manager creation if caching is enabled.

start_pokealarm.py
```diff
@@ -191,19 +208,37 @@ def parse_settings(root_path):
...
             log.critical("Names of Manager processes must be unique (regardless of capitalization)! Process will exit.")
-            sys.exit(1)
+            sys.exit(-1)
     log.info("Starting up the Managers")
     for m_name in managers:
         managers[m_name].start()
 
 
+def signal_handler(signal, frame):
+    print("CTRL-C catched, exiting gracefully...")
```
kvangent: Is this supposed to be "caught"?

start_pokealarm.py
```diff
@@ -85,14 +94,20 @@ def start_server():
...
-    spawn(manage_webhook_data, data_queue)
+    global webhook_manager
+    webhook_manager = gevent.spawn(manage_webhook_data, data_queue)
+
+    # make sure we stop properly
+    signal.signal(signal.SIGINT, signal_handler)
 
     # Start up Server
     log.info("PokeAlarm is listening for webhooks on: http://{}:{}".format(config['HOST'], config['PORT']))
+    global server
```
kvangent: This should go to the top of the file.

start_pokealarm.py
```diff
@@ -85,14 +94,20 @@ def start_server():
     logging.getLogger('connectionpool').setLevel(logging.WARNING)
     logging.getLogger('gipc').setLevel(logging.WARNING)
     logging.getLogger("urllib3").setLevel(logging.WARNING)
+    logging.getLogger('Cache').setLevel(logging.INFO)
 
     parse_settings(os.path.abspath(os.path.dirname(__file__)))
 
     # Start Webhook Manager in a Thread
-    spawn(manage_webhook_data, data_queue)
+    global webhook_manager
```
kvangent: This should go to the top of the file.

start_pokealarm.py
```diff
@@ -85,14 +94,20 @@ def start_server():
     logging.getLogger('connectionpool').setLevel(logging.WARNING)
     logging.getLogger('gipc').setLevel(logging.WARNING)
     logging.getLogger("urllib3").setLevel(logging.WARNING)
+    logging.getLogger('Cache').setLevel(logging.INFO)
```
kvangent: Is the default not INFO?

PokeAlarm/Manager.py
```diff
@@ -928,6 +956,11 @@ def process_raid(self, raid):
                 log.info("Raid {} ignored. Only {} seconds left.".format(gym_id, seconds_left))
             return
 
+        # don't alert about expired raids
+        if datetime.utcnow() > raid_end:
+            log.debug("Raid {} has expired, shame on whoever sent this old stuff")
```
kvangent: The guy who approved the PR didn't catch it either!

PokeAlarm/Manager.py
```diff
@@ -872,13 +901,12 @@ def process_egg(self, egg):
...
         start_time_str = get_time_as_str(egg['raid_begin'], self.__timezone)
-
-        gym_info = self.__gym_info.get(gym_id, {})
+        gym_info = self.get_gym_details(gym_id)
 
         egg.update({
-            "gym_name": gym_info.get('name', 'unknown'),
-            "gym_description": gym_info.get('description', 'unknown'),
-            "gym_url": gym_info.get('url', 'https://raw.githubusercontent.com/RocketMap/PokeAlarm/master/icons/gym_0.png'),
+            "gym_name": gym_info.get('name'),
```
kvangent: Why were the defaults removed?

kvangent:
On a meta note, what is the advantage of saving via pickle-ing instead of a more human readable format such as JSON as is done in the gym_cache branch?

PS. The graceful shutdown is very nice as well, good work on that.

PokeAlarm/Manager.py
fosJoddie: the defaults are set in the function that gets gym_info, in part because in the future we should not use github files as a source for these pictures. (much like the !pickle command in RM replaces the original files)

start_pokealarm.py
fosJoddie: I probably forgot to remove this after testing

fosJoddie:
for the `global` keyword: What is the correct way to initialize these on top of the file, for later initialization further down? I'm struggling to figure out exactly this, and feel like the "better" way to do this would be to refactor the methods and properties into a class

For the CTRL+C catching: this code catches the interrupt and flags the managers to start shutting down, if not done like this, the manager processes might end up in "limbo" where they t still has a running process even though they are not doing any work anymore. 

Lastly, I used the pickle because I felt like the built in functionality made it easiest to implement, and it would be files that are not going to be touched anyhow. (Arguably you could touch the gym files but they would be populated/overwritten soon anyhow). I assume that the pickle code is highly optimized compared to some random json write implementation.

kvangent:
My understanding of the `global` command is that it provides scope throughout the entire file (where as normally it is function wide scope). I prefer it at the top of the file because it makes it more clear that variable it a global value when trying to read the code.

Have you tested the graceful shutdown on Windows or OSX? Once we get into process level stuff OS matters.

One thing that I like about writing as JSON to a file is that users can add in their own gym_data if they prefer.

PokeAlarm/Manager.py
kvangent: What does the !pickle command in RM do?

fosJoddie:
I see I didn't write down my test environment. My local testing has been on Windows. 

My issue with making the saving as json is that it adds a complexity into the project that 99% of users will never need to utilize. Gyms change dynamically, and areas scanned also change, I see little to no benefit for a user to 100% control this himself. 

A user can easily write a tool himself to put things into the pickles though, if they so prefer.

PokeAlarm/Manager.py
fosJoddie: sorry, I meant to type `!sprite` - it's a custom discord command that gives you a download link to custom sprites to replace the original ones that are just numbers

kvangent:
@fosJoddie Is this ready to be reviewed again?

wfcook:
Does this allow the user to specify the cache file name? That would be essential for running multiple instances of PA

kvangent:
Closing this due to lack of response. 

I believe #465 implements almost everything included in this PR. @fosJoddie feel free to scope it out and let me know if I missed anything.

PR #450, opened 2017-10-11
Add new DTS for correct (EU) address formatting

fosJoddie:
## Description
Address DTS has been changed a tiny bit. It now will return an empty value if street name or street number is unknown. This allows for streets that have no numbers (like squares) to show up correctly. 

Example: A pokemon in a "Awesome Square, Europe City" would show up as "Awesome Square ???, Europe" it will now show up correctly with no street number.
Likewise unnamed roads wont show up as "unknown" but simply an empty string (there are many such roads in my area)

Lastly, and most importantly: there's a new DTS called `<address_eu>` that will write out "Alphabet Street 123" as addresses is usually written in Europe/Most of the world(?)

## Type of Change

## Motivation and Context
Two reasons: 
- Addresses in the EU is not written like in the US where street numbers come first. I would make the default address DTS EU style, but for historical reason I guess it is better to introduce `address-eu` as a new shorthand
- Seems counter intuitive to proclaim road numbers as ??? simply because google has no number information on them. Likewise unnamed streets should not be named "unkown".

## How Has This Been Tested?
Tested locally with an area with multiple unnumbered streets and unnamed streets

## Wiki Update

## Checklist

### DIFF
--- a/PokeAlarm/LocationServices.py
+++ b/PokeAlarm/LocationServices.py
@@ -79,9 +79,13 @@ class LocationService(object):
             for item in result['address_components']:
                 for category in item['types']:
                     loc[category] = item['short_name']
-            details['street_num'] = loc.get('street_number', '???')
-            details['street'] = loc.get('route', 'unknown')
+
+            # Note: for addresses in squares and on unnamed roads, it's correct with blank numbers/streetnames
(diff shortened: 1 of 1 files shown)


kvangent:
Looks good to me - just need wiki updates.

kvangent:
Normally my procedure is to wait for wiki updates but I want to get this out of the way.

PR #451, opened 2017-10-13
add team name to raid alert

ngovil21:
## Description
Add raid team name to raid alerts.

## Type of Change

## Motivation and Context

This gives more information on the current team controlling the raid. This information is useful because premiere balls are given based on the team controlling the raid. Gym team information is also extracted and saved earlier if gym notifications have been disabled.

## How Has This Been Tested?
This has been tested in a production environment with Discord notifications.

image

## Wiki Update

## Checklist

### DIFF
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -707,13 +707,17 @@ class Manager(object):
                 "url": gym['url']
             }
 
-        if self.__gym_settings['enabled'] is False:
-            log.debug("Gym ignored: notifications are disabled.")
-            return
-
         # Extract some basic information
         to_team_id = gym['new_team_id']
         from_team_id = self.__gym_hist.get(gym_id)
+
+        if from_team_id != to_team_id:
+            # Update gym's last known team
+            self.__gym_hist[gym_id] = to_team_id
+
+        if self.__gym_settings['enabled'] is False:
+            log.debug("Gym ignored: notifications are disabled.")
+            return
 
         # Doesn't look like anything to me
         if to_team_id == from_team_id:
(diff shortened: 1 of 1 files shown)


PokeAlarm/Manager.py
```diff
@@ -984,6 +986,9 @@ def process_raid(self, raid):
 
         gym_info = self.__gym_info.get(gym_id, {})
 
+        #team id saved in self.__gym_hist when processing gym
+        team_id = self.__gym_hist.get(gym_id, -1)
```
kvangent: This should default to '?' to be consistent with other DTS.

PokeAlarm/Manager.py
```diff
@@ -999,7 +1004,8 @@ def process_raid(self, raid):
             'dir': get_cardinal_dir([lat, lng], self.__location),
             'quick_move': self.__locale.get_move_name(quick_id),
             'charge_move': self.__locale.get_move_name(charge_id),
-            'form': self.__locale.get_form_name(pkmn_id, raid_pkmn['form_id'])
+            'form': self.__locale.get_form_name(pkmn_id, raid_pkmn['form_id']),
+            'team': self.__locale.get_team_name(team_id)
```
kvangent: Can we change `<team?` to `<team_name>` and then make sure we include `<team_id>` as well? It may be useful for URLs.

kvangent:
Also this will need Wiki support as well.

ngovil21:
Changed as requested.

Wiki update PR here: https://github.com/RocketMap/PokeAlarmWiki/pull/56

PR #462, opened 2017-10-21
remove "unknown" replacement for <form>, when the Pokemon has no form at all

JapuDCret:
## Description

The replacement "unknown" for <form> is annoying for Pokemon like Charizard that have no form.
You can fix this by creating a second alert, but this is way overkill and not a real solution.

## Type of Change

## Motivation and Context

This change is purely visual, but for my community it adds a lot.

## How Has This Been Tested?

I tested this with my local map for multiple days and fed some fake data to the PokéAlarm webhook(resulting in screenshots provided below)

Without this PR(using `"title":"<pkmn>-<form_id> <iv>% (<atk>,<def>,<sta>)"`)

With PR(using `"title":"<pkmn> <form> <iv>% (<atk>,<def>,<sta>) spawned"`)

## Wiki Update

## Checklist


kvangent:
The reason we put the DTS as "unknown" is because we've had issues with users in the past being able to tell if the DTS was working.

One solution I've talked about (and now implemented: b64c3eec4ac8d747db26b14756ef71e5b2820203) in the past was an `form_or_empty` DTS that would return empty instead of null. Do you feel that this satisfies your use case?

PR #465, opened 2017-11-04
Adding Caching 

kvangent:
## Description

This PR will switch PA to using `Cache` objects to store information that is not present on runtime. You can choose your cache type with `-ct` or `--cache-type`. The default cache type is `mem` which is a memory only - information will be lost between sessions. The second type of cache is `file` which will back up all stored information to a binary file located in the `cache` folder. This cache is backed up once a minute, and before PA exits.

File caches can be cleared by deleting the `cache/<manager_name>.cache` file. 

This also makes it easy to implement different kinds of cache's in the future, such as with a database. 

## Type of Change

## Motivation and Context

This will allow people to make changes to their config and restart PA without worrying about double notifications or gym info.

## How Has This Been Tested?

Poorly.

Screenshot of code:
img

## Wiki Update

## Checklist


billyjbryant:
Is there a config equivalent of `-ct` or is it `cache-type`?

PR #468, opened 2017-11-05
Update filters.json.example - new raid bosses

OldShaterhan:
Updating filters.json.example due to new raid bosses

## Description
Starting today new raid bosses are available on raids. It's the reason to pull that request.

## Type of Change

## Motivation and Context
To supple lacks of new raid bosses

## How Has This Been Tested?
Local instance.

image

## Wiki Update

## Checklist


MARTINI1:
Not only are raid bosses added, some are also removed

Look at this so you can see which ones are left that are less than those that are leaving
https://www.reddit.com/r/TheSilphRoad/comments/7au3d4/new_raid_bosses_collection_thread/

and the ivysaur tabulate is incorrect, correct this

OldShaterhan:
I know that some were removed, but, we don't know Niantic plans and maybe it would be better to rather add all Pokemon to raid filter to avoid changes in future migrations?

kvangent:
Adding all Pokemon seems unnecessary - that's just another 700 lines in the file.

OldShaterhan:
@kvangent , on the other hand it would add possibility to doesn't making changes in future in filters.json,exmple when rotation will occur, because all would be ready and local PA instances will be working immediately on new raid bosses, without necessity to restart instances/updating file again.

kvangent:
@OldShaterhan I think for now the example should just have relevant tiers. If people are concerned about missing Pokemon, they can certainly copy the top half of the alarms. But I think it's overwhelming to most users.

OldShaterhan:
That's right, but IMHO it's just better to be prepared and some new lines of code nowadays doesn't have big matter ;)

kvangent:
We pulled in #477 which fixed filters.json.

PR #469, opened 2017-11-06
Updating Webhook standards

billyjbryant:
## Description
Updated Webhook-Standards with updated detail from RocketMap Webhooks. 

## Type of Change

## Motivation and Context
This is done to ensure that the project stays up to date with the data being provided by RocketMap

## How Has This Been Tested?
This was tested on my dev server

N/A

## Wiki Update

## Checklist


PokeAlarm/WebhookStructs.py
```diff
@@ -52,6 +52,10 @@ def pokemon(data):
             'id': data['encounter_id'],
             'pkmn_id': int(data['pokemon_id']),
             'disappear_time': datetime.utcfromtimestamp(data['disappear_time']),
+            'time_until_despawn': data['seconds_until_despawn'],
+            'spawn_start': data['spawn_start'],
```
kvangent: Does monocle have these fields as well?

billyjbryant: I will set up a monocle instance to verify

kvangent: If it doesn't, we'll need to switch these to use `.get()` and default to `unknown`

PR #471, opened 2017-11-07
Fix underscore problems in Cache.py

sLoPPydrive:
## Description
In Cache.py there were some discrepancies between variables with one or two underscores. I changed them all to two underscores.

## Type of Change

## Motivation and Context
Bugfixing...

## How Has This Been Tested?
Local instance

## Wiki Update

## Checklist


kvangent:
Fixed in https://github.com/PokeAlarm/PokeAlarm/commit/d43fe9f5a2278eb48ff4ba5d7c245dca40efdb79.

The attributes need to be a single underscore to prevent name mangling so that the FileCache can access Cache attributes.

PR #474, opened 2017-11-08
Eggs expire when the raid starts, not when it ends

sLoPPydrive:
## Description
In my opinion a cached egg webhook data should expire when the raid starts, not when it ends.

## Type of Change

## Motivation and Context
I <3 being correct!

## How Has This Been Tested?
Local map.

## Wiki Update

## Checklist

### DIFF
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -843,8 +843,6 @@ class Manager(object):
 
         gym_id = egg['id']
 
-        raid_end = egg['raid_end']
-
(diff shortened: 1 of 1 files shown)


kvangent:
I had suspected this was wrong, but fairly certain it was the way it was working before which confused me.

sLoPPydrive:
It's logical that it also works as it had before because when there is either an egg or a raid there cannot be another egg or raid on this gym, so it also works if the egg expires when the raid **ends**. But it's confusing in the code. ;-)

PR #478, opened 2017-11-09
Adding gym_name support for monocle on eggs/raids

Hydro74000:
## Type of Change

## Motivation and Context
Currently, gym names are not sent in notifications for monocle. Monocle sends the gym name in raid/egg data, and not with gym info.
This PR will add the gym names for monocle users, without any effect for RM users

## How Has This Been Tested?
Tested on my own server since 2 weeks (monocle) and did a basic test with a simple RM instance.

N/A

## Wiki Update

## Checklist

### DIFF
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -876,7 +876,7 @@ class Manager(object):
         gym_info = self.__cache.get_gym_info(gym_id)
 
         if egg['gym_name'] == 'unknown':  # if RM or empty
-            egg.update({"gym_name": gym_info['name']})
+            egg.['gym_name'] = gym_info['name']
 
         egg.update({
             "gym_description": gym_info['description'],
@@ -991,7 +991,7 @@ class Manager(object):
         min_cp, max_cp = get_pokemon_cp_range(pkmn_id, 20)
 
 	if raid['gym_name'] == 'unknown':  # if RM or empty
-	    raid.update({"gym_name": gym_info['name']})
+	    raid['gym_name'] = gym_info['name']
         
         raid.update({
             'pkmn': name,
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -875,8 +875,10 @@ class Manager(object):
         team_id = self.__cache.get_gym_team(gym_id)
         gym_info = self.__cache.get_gym_info(gym_id)
 
+        if egg['gym_name'] == 'unknown':  # if RM or empty
+            egg.update({"gym_name": gym_info['name']})
+
         egg.update({
-            "gym_name": gym_info['name'],
             "gym_description": gym_info['description'],
             "gym_url": gym_info['url'],
             'time_left': time_str[0],
@@ -988,6 +990,9 @@ class Manager(object):
         form = self.__locale.get_form_name(pkmn_id, form_id)
         min_cp, max_cp = get_pokemon_cp_range(pkmn_id, 20)
 
+	if raid['gym_name'] == 'unknown':  # if RM or empty
+	    raid.update({"gym_name": gym_info['name']})
+        
         raid.update({
             'pkmn': name,
             'pkmn_id_3': '{:03}'.format(pkmn_id),
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -876,7 +876,7 @@ class Manager(object):
         gym_info = self.__cache.get_gym_info(gym_id)
 
         if egg['gym_name'] == 'unknown':  # if RM or empty
-            egg.['gym_name'] = gym_info['name']
+            egg['gym_name'] = gym_info['name']
 
         egg.update({
             "gym_description": gym_info['description'],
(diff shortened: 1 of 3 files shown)


PokeAlarm/Manager.py
```diff
@@ -893,8 +893,10 @@ def process_egg(self, egg):
         team_id = self.__cache.get_gym_team(gym_id)
         gym_info = self.__cache.get_gym_info(gym_id)
 
+        if egg['gym_name'] == 'unknown':  # if RM or empty
+            egg.update({"gym_name": gym_info['name']})
```
kvangent: Update shouldn't be used for a single key value pair. The correct way would be `egg['name'] = gym_info['name']`.

PokeAlarm/WebhookStructs.py
```diff
@@ -176,9 +176,15 @@ def egg(data):
...
 
+        if 'gym_name' in data:  # monocle sends gym_name in egg data.
+            gymname = check_for_none(str, data.get('gym_name'), 'unknown')
+        else:
+            gymname = 'unknown'
+
         egg = {
             'type': 'egg',
             'id': id_,
+            'gym_name': gymname,
```
kvangent: The point of the `check_for_none` is so that an explicit check isn't required - it should let you get the contents and default to `unknown` if it doesn't exists.

So this could be one line on : `'gym_name': check_for_none(str, data.get('gym_name'), 'unknown')`

PokeAlarm/WebhookStructs.py
```diff
@@ -246,7 +258,7 @@ def raid(data):
 
         raid['gmaps'] = get_gmaps_link(raid['lat'], raid['lng'])
         raid['applemaps'] = get_applemaps_link(raid['lat'], raid['lng'])
-
+
```
kvangent: This is a tab - PA uses spaces

PokeAlarm/WebhookStructs.py
```diff
@@ -220,9 +226,15 @@ def raid(data):
...
 
+        if 'gym_name' in data:  # monocle sends gym_name in raid data.
+            gymname = check_for_none(str, data.get('gym_name'), 'unknown')
+        else:
+            gymname = 'unknown'
+
         raid = {
             'type': 'raid',
             'id': id_,
+            'gym_name' : gymname,
```
kvangent: same as above

PokeAlarm/Manager.py
```diff
@@ -1005,9 +1007,11 @@ def process_raid(self, raid):
         form = self.__locale.get_form_name(pkmn_id, raid_pkmn['form_id'])
         min_cp, max_cp = get_pokemon_cp_range(pkmn_id, 20)
 
+	if raid['gym_name'] == 'unknown':  # if RM or empty
```
kvangent: same as above

kvangent:
I left some short comments. 

This big issue with this PR is that it doesn't cache the results into the cache, which means it'll avoid file cache. Also, it doesn't deal with the other gym info.

Do you know what Monocles webhooks look like?

Hydro74000:
Monocle is not sending 'gym' and 'gym_details' to webhooks so PA gym cache stay empty (reason why gym_name doesn't appears).

Chrales fork (new "official" monocle since Noctem vanished a few moth ago) doesn't send eggs/raids thru webhooks. 
Most of the other forks does.

For them, gym_name, gym_id and gym_url are part of the 'raid' message :

From Monkey :

                "raid_seed: raid['external_id'],
                "latitude": fort['lat'],
                "longitude": fort['lon'],
                "level": raid['level'],
                "pokemon_id": raid['pokemon_id'],
                "team": fort['team'],
                "cp": raid['cp'],
                "move_1": raid['move_1'],
                "move_2": raid['move_2'],
                "raid_begin": raid['time_battle'],
                "raid_end": raid['time_end'],
                "gym_id": raid["fort_external_id"],
                "base64_gym_id": b64encode(raid['fort_external_id'].encode('utf-8')),
                "gym_name": gym_name,
                "gym_url": gym_url

I could be possible to build a light gym cache with these info, but i don't think it's the right way (most of gym details are missing). That's, why I didn't use the cache in this PR

fosJoddie:
There has been discussions (and PRs) to include the gym info directly in raid webhooks from RocketMap as well, so this support would be welcome.

But at the very least you need to add support for the basic gym info:

```
gym_name
gym_ur
gym_description
team or team_id
``` 

I propose you send raid/egg info into the function to get gym info and then use the info directly if provided, otherwise look up in cache as before. Caching of the info at that point would be pointless since the webhook structure implies the info is always provided (and probably cached at RM/Monocle side)

mali055:
https://github.com/Hydro74000/Monocle

I've been using the above fork, please see if you can get it to work.

kvangent:
@mali055 Hyrdo's updated his fork to send info through gym-info webhooks, so it should be working now. This PR should work with other versions of PA outta the box. We're also going to reach out to some of the other Monocle forks and create a standard so that everyone works out of the box.

Hydro74000:
@fosJoddie I added gym info processing from raids and eggs in my last commit.

I need a RM tester for these commits pls.

kvangent:
@Hydro74000 https://github.com/PokeAlarm/PokeAlarm/pull/519 features a massive event rework, hoping that it works in monocle too

kvangent:
Closing this as it's massively out of date and should hopefully not be needed after #519.

PR #480, opened 2017-11-10
Simple file logging via --log <filename>

sLoPPydrive:
## Description
If given `--log <filename>` on command line or in `config.ini` PokeAlarm logs everything you see on screen to the given log file. The log line format is the same as the screen output.

I also removed some whitespace from `config.ini.example` and replaced tabs with spaces for better readability.

## Type of Change

## Motivation and Context
In order to debug some things I added very simple logging to file for later investigations.

## How Has This Been Tested?
Local and remote installations

## Wiki Update

## Checklist

### DIFF
--- a/start_pokealarm.py
+++ b/start_pokealarm.py
@@ -19,7 +19,6 @@ import Queue
 import json
 import os
 import sys
-from time import strftime
 # 3rd Party Imports
 from flask import Flask, request, abort
 # Local Imports
@@ -174,25 +173,13 @@ def parse_settings(root_path):
     parser.add_argument(
         '-tz', '--timezone', type=str, action='append', default=[None],
         help='Timezone used for notifications. Ex: "America/Los_Angeles"')
-    parser.add_argument('--log_to_file',
-                        help=('Enable logging to file.'),
-                        action='store_true', default=False)
-    parser.add_argument('--log_path',
-                        help=('Defines directory to save log files to.'),
-                        default='logs/')
+    parser.add_argument(
+        '--log', help='Log output to given file.')
 
     args = parser.parse_args()
 
-    if args.log_to_file:
-        # Create directory for log files.
-        if not os.path.exists(args.log_path):
-            os.mkdir(args.log_path)
-        date = strftime('%Y%m%d_%H%M')
-        filename = os.path.join(args.log_path, '{}.log'.format(date))
-        fh = logging.FileHandler(filename)
-        fh.setFormatter(
-            logging.Formatter(fmt='%(asctime)s [%(processName)15.15s][%(name)10.10s][%(levelname)8.8s] %(message)s'))
-        logging.root.addHandler(fh)
+    if args.log:
+        configure_file_logging(args.log)
 
     if args.debug:
         log.setLevel(logging.DEBUG)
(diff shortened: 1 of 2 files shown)


start_pokealarm.py
```diff
@@ -140,9 +141,26 @@ def parse_settings(root_path):
                         help='Maximum number of attempts an alarm makes to send a notification.')
     parser.add_argument('-tz', '--timezone', type=str, action='append', default=[None],
                         help='Timezone used for notifications.  Ex: "America/Los_Angeles"')
+    parser.add_argument('--log_to_file',
+                        help=('Enable logging to file.'),
+                        action='store_true', default=False)
+    parser.add_argument('--log_path',
+                        help=('Defines directory to save log files to.'),
+                        default='logs/')
```
billyjbryant: Please add a short option to both of these config parameters.  

Also please ensure that the argument type, action, and default value are being specified to conform with coding standards.

start_pokealarm.py
```diff
@@ -140,9 +141,26 @@ def parse_settings(root_path):
...
+        # Create directory for log files.
+        if not os.path.exists(args.log_path):
+            os.mkdir(args.log_path)
+        date = strftime('%Y%m%d_%H%M')
+        filename = os.path.join(args.log_path, '{}.log'.format(date))
+        fh = logging.FileHandler(filename)
+        fh.setFormatter(
+            logging.Formatter(fmt='%(asctime)s [%(processName)15.15s][%(name)10.10s][%(levelname)8.8s] %(message)s'))
+        logging.root.addHandler(fh)
+
```
billyjbryant: Have you tested potentially adding the ability to control the verbosity of resulting logs, such as giving the option to only log ERRORS, etc?

start_pokealarm.py
kvangent: +1, short flags are good.

I feel like two flags feels redundant: is it possible to combine them into a single flag somehow? Just a suggestion.

start_pokealarm.py
kvangent: Probably beyond the scope of this for now.

start_pokealarm.py
```diff
@@ -140,9 +141,26 @@ def parse_settings(root_path):
...
+                        default='logs/')
 
     args = parser.parse_args()
 
+    if args.log_to_file:
+        # Create directory for log files.
+        if not os.path.exists(args.log_path):
+            os.mkdir(args.log_path)
+        date = strftime('%Y%m%d_%H%M')
+        filename = os.path.join(args.log_path, '{}.log'.format(date))
```
kvangent: Any particular reason you decided to log to a new file each time instead of just appending?

start_pokealarm.py
```diff
@@ -140,9 +141,26 @@ def parse_settings(root_path):
...
+    parser.add_argument('--log_path',
+                        help=('Defines directory to save log files to.'),
+                        default='logs/')
 
     args = parser.parse_args()
 
+    if args.log_to_file:
+        # Create directory for log files.
+        if not os.path.exists(args.log_path):
+            os.mkdir(args.log_path)
```
kvangent: Is this recursive? What happens if the user doesn't have write access?

mali055:
Could this log debug messages as well, when -d switch is active.

kvangent:
In theory it should, but we've had a couple of users reporting in that it doesn't always capture logs from managers.

LegitDongo:
I think that only one flag is really necessary since you can include both the path and the name of the log all in one string.
This is how most programs I've seen do it
Ex. `--log /opt/pa/logs/mylogfile1.log`
This makes more sense logically because it could get confusing, for example, if you should include a `/` at the end of the path in the `--log_path` parameter or not

start_pokealarm.py
sLoPPydrive: To be honest I simply copied over the functionality from RocketMap (2 commandline params + new file for each launch).

start_pokealarm.py
sLoPPydrive: When I combine both parameters to something like `--log /some/dir/pokealarm.log` it will be hard to find a non-ridiculous short option since `-l` is already taken and everything else just doesn't feel right. Also I think that `--log` would be short enough. What do you think?

kvangent:
How does everyone feel about `-lf` and `--log-file`?

sLoPPydrive:
That's okay for me. :) And actually it's more correct than just `--log` since you are directly specifying a file.

kvangent:
Closing this since it is significantly out of date and has issues.

PR #484, opened 2017-11-12
Webhook Tester Location Fix

### LINKED ISSUE #483: [BUG] Webhook Tester Location
## Troubleshooting Checklist
- [x] My RocketMap install is at the most recent version.  
- [x] I've already checked the
 [FAQ](https://github.com/RocketMap/PokeAlarm/wiki/faq) and the
 [Wiki](https://github.com/RocketMap/PokeAlarm/wiki) already.  

## Problem Description
Location doesn't work
Some description text should be updated

## Environment
Operating System: N/A
Python Version: N/A
PokeAlarm Branch: master
PokeAlarm Version: N/A

## PokeAlarm Setup:
Your config file: N/A
Your filters file: N/A
Your alarms file:  N/A

## Log Output
N/A

LegitDongo:
## Description
Fixes issue with location not getting set
Also a few minor fixes
Will fix #483

## Type of Change

## Motivation and Context
A lot of people have geofences set up with their PA, and without location settings being fixed the webhook tester will always show as outside of geofence

## How Has This Been Tested?
Tested locally by using it

## Wiki Update

## Checklist

### DIFF
--- a/tools/webhook_test.py
+++ b/tools/webhook_test.py
@@ -3,7 +3,6 @@ import time
 import sys
 import json
 import re
-from random import randint
 
 truthy = frozenset([
     "yes", "Yes", "y", "Y", "true", "True", "TRUE", "YES", "1", "!0"
@@ -29,6 +28,7 @@ teams_formatted = re.sub('[{}",]', '', json.dumps(teams, indent=4, sort_keys=Tru
 
 def set_init(webhook_type):
     payloadr = {}
+    current_time = time.time()
     if webhook_type == whtypes["1"]:
         payloadr = {
             "type": "pokemon",
@@ -36,9 +36,9 @@ def set_init(webhook_type):
                 "pokemon_id": 149,
                 "pokemon_level": 30,
                 "player_level": 30,
-                "latitude": 33.980823,
-                "longitude": -81.052988,
-                "encounter_id": randint(1, 99999999),
+                "latitude": 37.7876146,
+                "longitude": -122.390624,
+                "encounter_id": current_time,
                 "cp_multiplier": 0.7317000031471252,
                 "form": 15,
                 "cp": 768,
@@ -60,10 +60,10 @@ def set_init(webhook_type):
         payloadr = {
             "type": "pokestop",
             "message": {
-                "pokestop_id": randint(1, 99999999),
+                "pokestop_id": current_time,
                 "enabled": "True",
-                "latitude": 62.790967,
-                "longitude":  76.927920,
+                "latitude": 37.7876146,
+                "longitude": -122.390624,
                 "active_fort_modifier": 0
             }
         }
@@ -81,8 +81,8 @@ def set_init(webhook_type):
                 "lowest_pokemon_motivation": 0.8795773983001709,
                 "total_cp": 11099,
                 "enabled": "True",
-                "latitude": 62.790967,
-                "longitude":  76.927920
+                "latitude": 37.7876146,
+                "longitude": -122.390624
             }
         }
     elif webhook_type == whtypes["4"]:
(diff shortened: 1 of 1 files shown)


tools/webhook_test.py
```diff
@@ -43,7 +38,7 @@ def set_init(webhook_type):
                 "player_level": 30,
                 "latitude": 33.980823,
                 "longitude": -81.052988,
-                "encounter_id": 0,
+                "encounter_id": randint(1, 99999999),
```
kvangent: Why don't we just make this time based and remove the chance to repeats entirely?

tools/webhook_test.py
```diff
@@ -171,23 +166,32 @@ def get_and_validate_team():
...
+print "What is the URL of where you would like to send the webhook? (default: http://127.0.0.1:4000)\n>",
 url = raw_input()
 if url == '' or url.isspace():
     url = "http://127.0.0.1:4000"
-    log.info("Assuming " + url + " as webhook URL")
+    print "Assuming " + url + " as webhook URL"
 
 print "Does location matter?\n>",
 if raw_input() in truthy:
+    regex_coords = re.compile("^[-+]?[0-9]*\.?[0-9]*$", re.MULTILINE)
```
kvangent: Should we consider asking for location in format `lat, lng`? Or do you think it's better as 2 questions?

LegitDongo: Either is fine. I was doing it this way because it was easy :smile:

tools/webhook_test.py
billyjbryant: I agree with this, if you truly want to make it fully random and unique, generate a hash off of the current epoch

tools/webhook_test.py
billyjbryant: I prefer the `lat,lng` option.  easier to copy and paste

tools/webhook_test.py
```diff
@@ -171,23 +166,32 @@ def get_and_validate_team():
...
 
-print "What is the URL of where you would like to send the webhook? (default: 127.0.0.1:4000)\n>",
+print "What is the URL of where you would like to send the webhook? (default: http://127.0.0.1:4000)\n>",
 url = raw_input()
 if url == '' or url.isspace():
     url = "http://127.0.0.1:4000"
-    log.info("Assuming " + url + " as webhook URL")
+    print "Assuming " + url + " as webhook URL"
 
 print "Does location matter?\n>",
```
billyjbryant: We should consider also asking in this question if a user uses geofences.  Because if they do, then location DOES matter whether they think it does or not.

tools/webhook_test.py
```diff
@@ -43,7 +38,7 @@ def set_init(webhook_type):
                 "player_level": 30,
                 "latitude": 33.980823,
                 "longitude": -81.052988,
```
billyjbryant: Consider using the defaults from the PA wiki: 37.7876146,-122.390624 for default location values.  :)

tools/webhook_test.py
```diff
@@ -235,22 +239,21 @@ def get_and_validate_team():
...
             raise requests.exceptions.RequestException(
                 "Response received {}, webhook not accepted.".format(resp.status_code))
-            log.info("Attempting connection again")
+            print "Attempting connection again"
     print "Send again?\n>",
     if raw_input() not in truthy:
         break
     if payload["type"] == "pokemon":
-        payload["message"]["encounter_id"] += 1
+        payload["message"]["encounter_id"] = randint(1, 9999999)
```
billyjbryant: Same concern as @kvangent's earlier suggestion.

tools/webhook_test.py
```diff
@@ -229,28 +233,23 @@ def get_and_validate_team():
...
-            log.info("Attempting connection again")
+            print "Attempting connection again"
     print "Send again?\n>",
     if raw_input() not in truthy:
         break
-    if payload["type"] == "pokemon":
-        payload["message"]["encounter_id"] += 1
-    elif payload["type"] == "gym":
+    if payload["type"] == "gym":
         print "Which gym? (put in number)" + teams_formatted + "\n>",
```
kvangent: Should this be which team?

PR #487, opened 2017-11-14
Correct incorrect cache help description

tomballgithub:
## Description

The help text for the new cache_type configuration has a cut/paste error which errantly mentions the google API.  This corrects that as well as making it clear that there are two choices, mem or file.

## Type of Change

## Motivation and Context

Help text should be correct

## Wiki Update

## Checklist

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -18,7 +18,7 @@
 #geofence:										# File containing geofence(s) used to filter (default: None)
 #location:										# Location for the manager. 'Name' or 'lat lng' (default: None)
 #locale:										# Language to be used to translate names (default: en)
-#cache_type:                                    # Type of cache to use: mem or file. (default: 'mem')
+#cache_type:                                    # Specify the type of cache to use. Options: ['mem', 'file'] (Default: 'mem')
 #unit:											# Units used to measure distance. Either 'imperial' or 'metric' (default: imperial)
 #timelimit:										# Minimum number of seconds remaining to send a notification (default: 0)
 #max_attempts:									# Maximum number of attempts an alarm makes to send a notification. (default: 3)
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -18,7 +18,7 @@
 #geofence:										# File containing geofence(s) used to filter (default: None)
 #location:										# Location for the manager. 'Name' or 'lat lng' (default: None)
 #locale:										# Language to be used to translate names (default: en)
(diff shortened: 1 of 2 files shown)


start_pokealarm.py
```diff
@@ -133,7 +133,7 @@ def parse_settings(root_path):
                         help='Specify either metric or imperial units to use for distance measurements. ')
     parser.add_argument('-ct', '--cache_type', type=parse_unicode, action='append', default=['mem'],
                         choices=cache_options,
-                        help='Specify a Google API Key to use.')
+                        help='Select cache storage location (mem or file)')
```
kvangent: Change this to `Specify the type of cache to use. Options: ['mem', 'file'] (Default: 'mem')`

config/config.ini.example
```diff
@@ -18,7 +18,7 @@
 #geofence:										# File containing geofence(s) used to filter (default: None)
 #location:										# Location for the manager. 'Name' or 'lat lng' (default: None)
 #locale:										# Language to be used to translate names (default: en)
-#cache_type:                                    # Type of cache to use. (default: 'mem')
+#cache_type:                                    # Type of cache to use: mem or file. (default: 'mem')
```
kvangent: Change this to `Specify the type of cache to use. Options: ['mem', 'file'] (Default: 'mem')`

tomballgithub:
I'll change it.  My feedback is that 'mem' and 'file' are not really documented anywhere.  That is why I tried to say this really boils down to the 'cache storage location'.

tomballgithub:
Changes implemented.

kvangent:
The reason I want to avoid saying 'storage location' is because in the future there could be a cache that operates different for whatever reason but still uses a file or mem.

As for documentation, it's mentioned both in [server settings](https://github.com/PokeAlarm/PokeAlarm/wiki/server-settings) and in depth in [Object Caching](https://github.com/PokeAlarm/PokeAlarm/wiki/Object-Caching)

tomballgithub:
It really needs to documented on the github site because the only way I could find the info was go to the discord #announcements and click on a link there to get to the 3.4 notes.  A normal person looking for the info at the PokeAlarm github will be challenged.

kvangent:
The wiki in integrated into the Github site. You can access it from the wiki tab at the top of the page.

tomballgithub:
Thanks, I found: https://github.com/PokeAlarm/PokeAlarm/wiki/Object-Caching

Issue #452, opened 2017-10-16
[ENHANCEMENT] Adding multiple number support to Twilio.


boby361:
## Description
Alter Twilio to allow multiple number support per worker. 

## Why is this enhancement necessary or useful?
For someone who may want to add a lot of numbers to the same fence and filter this would be significantly cleaner then running many many workers.
All values but the "to_number" are consistent and Twilio has support for multiple numbers being sent at once. It caches it and delays each thing sent to it by 1 second.

## How can this enhancement be accomplished?
Iv been told by multiple people in the community that this is a very easy thing to implement. (Although no one seems to want to help)

kvangent:
Commit 633f27eee2905c7c8784cc6322161a8600a92c9d

billyjbryant:
This is merged and will be implemented in 3.5

kvangent:
(leaving this open until dev is pulled)

PR #491, opened 2017-11-18
Twilio Multi "To_Number" Support

billyjbryant:
## Description
This PR adds support for sending alerts to multiple numbers via a Twilio.  This works by replacing the  `to_number` field of the alarm setting with an array of numbers.  

## Type of Change

## Motivation and Context
This was a requested feature under https://github.com/PokeAlarm/PokeAlarm/issues/452.  The feature gives the ability to subscribe multiple recipients for PokeAlarm alerts through Twilio. 

## How Has This Been Tested?
This was tested through my trial Twilio account using both a single number destination and an array of up to 3 individual numbers.  Tests were performed using the webhook_test utility. 

N/A

## Wiki Update

The wiki will need to be updated to indicate the changes to use the alert appropriately.  

## Checklist

### DIFF
--- a/PokeAlarm/Twilio/TwilioAlarm.py
+++ b/PokeAlarm/Twilio/TwilioAlarm.py
@@ -74,7 +74,7 @@ class TwilioAlarm(Alarm):
                 to_num=self.__to_number,
                 from_num=self.__from_number,
                 body="PokeAlarm activated!"
-                )
+            )
             log.info("Startup message sent!")
 
     # Set the appropriate settings for each alert
@@ -89,11 +89,11 @@ class TwilioAlarm(Alarm):
 
     # Send Pokemon Info
     def send_alert(self, alert, info):
-            self.send_sms(
-                to_num=alert['to_number'],
-                from_num=alert['from_number'],
-                body=replace(alert['message'], info)
-            )
+        self.send_sms(
+            to_num=alert['to_number'],
+            from_num=alert['from_number'],
+            body=replace(alert['message'], info)
+        )
 
     # Trigger an alert based on Pokemon info
     def pokemon_alert(self, pokemon_info):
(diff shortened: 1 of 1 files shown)


PokeAlarm/Twilio/TwilioAlarm.py
```diff
@@ -74,7 +74,7 @@ def startup_message(self):
                 to_num=self.__to_number,
                 from_num=self.__from_number,
                 body="PokeAlarm activated!"
-            )
+                )
```
kvangent: I don't think you need this indent

PokeAlarm/Twilio/TwilioAlarm.py
```diff
@@ -89,11 +89,11 @@ def set_alert(self, settings, default):
 
     # Send Pokemon Info
     def send_alert(self, alert, info):
-        self.send_sms(
-            to_num=alert['to_number'],
-            from_num=alert['from_number'],
-            body=replace(alert['message'], info)
-        )
+            self.send_sms(
```
kvangent: too many indents here too

PokeAlarm/Twilio/TwilioAlarm.py
```diff
@@ -117,9 +117,18 @@ def raid_alert(self, raid_info):
 
     # Send a SMS message
     def send_sms(self, to_num, from_num, body):
-        args = {
-            'to': to_num,
-            'from_': from_num,
-            'body': body
-        }
-        try_sending(log, self.connect, "Twilio", self.__client.messages.create, args)
+        if isinstance(to_num, list):
```
kvangent: Could we reduce this? Instead of branching, can we do something like:
```python
if not isinstance(to_num, list):
     to_num = [to_num]
for num in to_num:
```
We could also do the check when we set in the number so we don't have to redo it for each message send.

PokeAlarm/Twilio/TwilioAlarm.py
billyjbryant: I'll fix.

billyjbryant:
Updated per request.

tomballgithub:
I love this idea!

PR #492, opened 2017-11-18
Added 'gymname_contains' to gym filters.

tomballgithub:
## Description

This adds the ability to format an egg or raid alert to only occur when at a sponsored gym.  Sponsored gyms are identified by substrings provided via config file or command line such as [sprint,starbucks,level badge,phone skin]

Add the optional 'sponsored_raid' tag in the Egg and/or Raid sections of a format file.  For those alerts, the gym name will be checked to see if it includes the substrings provided earlier that indicate a sponsored gym.  It there is a match, the alert is allowed otherwise it is blocked.

## Type of Change

## Motivation and Context

Gives you the ability to alert on raids at sponsored gyms

## How Has This Been Tested?

Tested on Ubuntu 16.04 in a production setting. 

## Wiki Update

## Checklist

### DIFF
--- a/PokeAlarm/Filters.py
+++ b/PokeAlarm/Filters.py
@@ -155,6 +155,10 @@ def load_egg_section(settings):
         "contains": settings.pop('gymname_contains', [])
     }
 
+    if not isinstance(egg['contains'], list):
+            log.error("'gymname_contains' filter must be a list")
+            sys.exit(1)
+
     log.debug("Report eggs between level {} and {}".format(
         egg['min_level'], egg['max_level']))
 
@@ -169,6 +173,10 @@ def load_raid_section(settings):
         "contains": settings.pop('gymname_contains', [])
     }
 
+    if not isinstance(raid['contains'], list):
+            log.error("'gymname_contains' filter must be a list")
+            sys.exit(1)
+    
     # load any raid pokemon filters
     filters = load_pokemon_filters(settings)
     raid['filters'] = filters
--- a/PokeAlarm/Filters.py
+++ b/PokeAlarm/Filters.py
@@ -151,7 +151,8 @@ def load_egg_section(settings):
     egg = {
         "enabled": bool(parse_boolean(settings.pop('enabled', None)) or False),
         "min_level": int(settings.pop('min_level', 0) or 0),
-        "max_level": int(settings.pop('max_level', 10) or 10)
+        "max_level": int(settings.pop('max_level', 10) or 10),
+        "contains": settings.pop('gymname_contains', [])
     }
 
     log.debug("Report eggs between level {} and {}".format(
@@ -164,7 +165,8 @@ def load_egg_section(settings):
 def load_raid_section(settings):
     log.info("Setting up Raid Filters...")
     raid = {
-        "enabled": bool(parse_boolean(settings.pop('enabled', None)) or False)
+        "enabled": bool(parse_boolean(settings.pop('enabled', None)) or False),
+        "contains": settings.pop('gymname_contains', [])
     }
 
     # load any raid pokemon filters
--- a/PokeAlarm/Filters.py
+++ b/PokeAlarm/Filters.py
@@ -176,7 +176,7 @@ def load_raid_section(settings):
     if not isinstance(raid['contains'], list):
             log.error("'gymname_contains' filter must be a list")
             sys.exit(1)
-    
+
     # load any raid pokemon filters
     filters = load_pokemon_filters(settings)
     raid['filters'] = filters
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -956,6 +956,13 @@ class Manager(object):
             return
 
         gym_id = egg['id']
+        gym_info = self.__cache.get_gym_info(gym_id)
+        gym_name = gym_info['name'].lower()
+
+        # Check if egg gym should be sponsored and is sponsored
+        if (self.__egg_settings['sponsored_raid'] is True and not any(x in gym_name for x in config['SPONSORED_GYMS'])):
+            log.debug("Egg {} is not at a sponsored gym: ".format(gym_name))
+            return
 
         # Check if egg has been processed yet
         if self.__cache.get_egg_expiration(gym_id) is not None:
@@ -1009,7 +1016,6 @@ class Manager(object):
 
         # team id saved in self.__gym_hist when processing gym
         team_id = self.__cache.get_gym_team(gym_id)
-        gym_info = self.__cache.get_gym_info(gym_id)
 
         egg.update({
             "gym_name": gym_info['name'],
@@ -1044,9 +1050,16 @@ class Manager(object):
             return
 
         gym_id = raid['id']
+        gym_info = self.__cache.get_gym_info(gym_id)
+        gym_name = gym_info['name'].lower()
 
         pkmn_id = raid['pkmn_id']
         raid_end = raid['raid_end']
+
+        # Check if raid is sponsored
+        if (self.__raid_settings['sponsored_raid'] is True and not any(x in gym_name for x in config['SPONSORED_GYMS'])):
+            log.debug("Raid {} is not at a sponsored gym: ".format(gym_name))
+            return
 
         # Check if raid has been processed
         if self.__cache.get_raid_expiration(gym_id) is not None:
@@ -1118,8 +1131,15 @@ class Manager(object):
                 self.__location, [lat, lng], raid)
 
         if self.__quiet is False:
+<<<<<<< HEAD
             log.info("Raid ({}) notification ".format(gym_id)
                      + "has been triggered!")
+=======
+            if (self.__raid_settings['sponsored_raid'] is True):
+                log.info("Sponsored Raid ({}) notification has been triggered!".format(gym_id))
+            else:
+                log.info("Raid ({}) notification has been triggered!".format(gym_id))
+>>>>>>> d22f0d7... Sponsored Raids
 
         time_str = get_time_as_str(
             raid['raid_end'], self.__timezone)
@@ -1127,7 +1147,6 @@ class Manager(object):
 
         # team id saved in self.__gym_hist when processing gym
         team_id = self.__cache.get_gym_team(gym_id)
-        gym_info = self.__cache.get_gym_info(gym_id)
         form_id = raid_pkmn['form_id']
         form = self.__locale.get_form_name(pkmn_id, form_id)
         min_cp, max_cp = get_pokemon_cp_range(pkmn_id, 20)
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -985,7 +985,7 @@ class Manager(object):
                        for x in self.__egg_settings['contains']):
                 log.info("Egg {} ignored: gym name did not match the "
                          "gymname_contains "
-                         "filter.".format(gym_info['name'].lower()))
+                         "filter.".format(gym_id))
                 return
 
         # Check if raid is in geofences
@@ -1084,7 +1084,7 @@ class Manager(object):
                        for x in self.__raid_settings['contains']):
                 log.info("Raid {} ignored: gym name did not match the "
                          "gymname_contains "
-                         "filter.".format(gym_info['name'].lower()))
+                         "filter.".format(gym_id))
                 return
 
         # Check if raid is in geofences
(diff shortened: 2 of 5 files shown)


tomballgithub:
Debugging an issue...

kvangent:
There are a few implementation issues with this, but I feel like it might be more useful as a general "contains" field. Instead of a 'sponsored_gyms' to true or false, you could specify the list of substrings to check for in the filter themselves. It would look something like this:
```
"contains": ["sprint", "starbucks"]
```
What are your thoughts on that?

tomballgithub:
My concern is I have 80+ filter files and it would be easier to change this in one place for sponsored_gyms than trying to touch it in all those files, even with a script.  For instance sprint gyms have had 3 names over the last few months in my area, 'sprint', ''get your level badge', and 'free team phone skin'

billyjbryant:
I agree with @kvangent's suggestion, having this be a filter to allow changing what text is contained within a gym name would be a better implementation.

tomballgithub:
I agree with that also if you want the flexibility of different trigger words for different managers, which is a problem not in search of a solution right now.  It works great if you have 1 or two alarms but it doesn't scale very well.

I was trying to create an easy way to handle a current need requested by many, which is a common set of trigger points easily managed across all managers.

kvangent:
So my general thought on this is that Managers are supposed to be replacements for standalone instances, but with this addition a user would require launching multiple instances. So I think we are unwilling to implement this in it's current form, but I'm going to leave it open for now for any users that are interested to check it out. 

I think you're real problem stems from having to have 80+ managers to satisfy your configuration needs, which I'm hoping will no longer be case in Patch 3.5.

tomballgithub:
I'll say this as clearly as possible - I cannot argue with your point 'that Managers are supposed to be replacements for standalone instances, but with this addition a user would require launching multiple instances [if they wanted to check for different gym names]'.  :-)  I will look at improving it

tomballgithub:
Requested change has been made and tested indirectly.  Will test under real world conditions in morning once raids start

billyjbryant:
@tomballgithub you can also use the webhook-tester in the `tools` folder to test your alerts.

tomballgithub:
Duh!  Didn't think of that.  Thank you...

PokeAlarm/Filters.py
```diff
@@ -140,7 +140,8 @@ def load_egg_section(settings):
     egg = {
         "enabled": bool(parse_boolean(settings.pop('enabled', None)) or False),
         "min_level": int(settings.pop('min_level', 0) or 0),
-        "max_level": int(settings.pop('max_level', 10) or 10)
+        "max_level": int(settings.pop('max_level', 10) or 10),
+        "contains": settings.pop('gymname_contains', [])
```
kvangent: We should probably do some check to ensure this is a list. If it's a string bad things will happen.

PokeAlarm/Filters.py
```diff
@@ -152,7 +153,8 @@ def load_egg_section(settings):
 def load_raid_section(settings):
     log.info("Setting up Raid Filters...")
     raid = {
-        "enabled": bool(parse_boolean(settings.pop('enabled', None)) or False)
+        "enabled": bool(parse_boolean(settings.pop('enabled', None)) or False),
+        "contains": settings.pop('gymname_contains', [])
```
kvangent: list check

PokeAlarm/Manager.py
```diff
@@ -847,6 +847,14 @@ def process_egg(self, egg):
             return
 
         gym_id = egg['id']
+        gym_info = self.__cache.get_gym_info(gym_id)
+        gym_name = gym_info['name'].lower()
+
+        # Check if egg gym should be sponsored and is sponsored
+        if len(self.__egg_settings['contains']) > 0:
```
kvangent: This should be after processed check because we don't want to keep getting notified about the same egg getting rejected over and over - I would suggest moving it to right before geofences.

PokeAlarm/Manager.py
```diff
@@ -887,14 +895,16 @@ def process_egg(self, egg):
             self.__loc_service.add_optional_arguments(self.__location, [lat, lng], egg)
 
         if self.__quiet is False:
-            log.info("Egg ({}) notification has been triggered!".format(gym_id))
+            if len(self.__egg_settings['contains']) > 0:
+                log.info("Sponsored Egg ({}) notification has been triggered!".format(gym_id))
```
kvangent: No reason to differentiate these anymore.

PokeAlarm/Manager.py
```diff
@@ -847,6 +847,14 @@ def process_egg(self, egg):
             return
 
         gym_id = egg['id']
+        gym_info = self.__cache.get_gym_info(gym_id)
+        gym_name = gym_info['name'].lower()
```
kvangent: Instead of making this a variable, let's just use the actual `gym_info['name'].lower()` in the contains check. This will prevent anyone from accidentally using it and getting a name in lower case.

PokeAlarm/Manager.py
```diff
@@ -928,10 +938,18 @@ def process_raid(self, raid):
             return
 
         gym_id = raid['id']
+        gym_info = self.__cache.get_gym_info(gym_id)
+        gym_name = gym_info['name'].lower()
```
kvangent: same as above - (re)move to contains line

PokeAlarm/Manager.py
```diff
@@ -928,10 +938,18 @@ def process_raid(self, raid):
...
 
         pkmn_id = raid['pkmn_id']
         raid_end = raid['raid_end']
 
+        # Check if raid gym should be sponsored and is sponsored
+        if len(self.__raid_settings['contains']) > 0:
```
kvangent: same as above - move lower

PokeAlarm/Manager.py
```diff
@@ -997,14 +1015,16 @@ def process_raid(self, raid):
             self.__loc_service.add_optional_arguments(self.__location, [lat, lng], raid)
 
         if self.__quiet is False:
-            log.info("Raid ({}) notification has been triggered!".format(gym_id))
+            if len(self.__raid_settings['contains']) > 0:
+                log.info("Sponsored Raid ({}) notification has been triggered!".format(gym_id))
```
kvangent: Same as above - no reason to call out sponsored

filters.json.example
```diff
@@ -426,11 +426,13 @@
     },
     "eggs":{
         "enabled":"True",
+        "gymname_contains":["sprint","level badge","phone skin","starbucks"],
```
kvangent: I feel like this is better as `"contains"` but I'm interested to hearing why you decided on gymname_contains.

tomballgithub: I used gymname_contains because it was crystal clear.  'Contains' by itself could mean other things such as a pokemon in the gym or a trainer in the gym or the type of pokemon for the raid or the street address or some other attribute.  I'd prefer something more succinct but I erred on the side of specificity in the code that was user-visible.  I think someone looking at a JSON that has this section filled out will know what 'contains' means but otherwise it requires looking at the wiki to be sure.

PokeAlarm/Filters.py
tomballgithub: Can you recommend if/how I could do this with 'create_multi_filter' as I cannot figure that out?  Or should I just just create a separate function for that?

tomballgithub:
Made requested changes except I have a question on the list checking and if create_multi_filter can be used.  Made comment on your review.

PokeAlarm/Filters.py
kvangent: I don't think you should use create_mutli_filter. The safest bet is to just check if it's a list and raise a ValueException afterwards.

kvangent:
Please remember you'll need to rebase to dev as well.

tomballgithub:
Added check to ensure lists are provided otherwise do sys.exit(1). 
Rebased...

kvangent:
Just to be clear, you have to rebase to the `dev` branch to pass the TravisCI check.

tomballgithub:
I just copied to a new branch SponsoredRaids4 and rebased to 'dev'.  Please check the results.  Looks like it is picking up a lot of other stuff.  I do not have the expertise to understand what is going on.  Any advice?

I did the following and fixed the conflicts after each iteration of the rebasing
```
git remote add upstream https://github.com/PokeAlarm/PokeAlarm.git
git fetch upstream
git rebase upstream/dev
git push -f origin SponsoredRaids4
```

billyjbryant:
@tomballgithub try using `rebase -i upstream/dev` and commenting out the commits that are not related to your PR.

kvangent:
Additionally, when you rebase a branch you are rewriting it's history. You'll have to use `git push <remote_name> <branch_name> -f` to 'force' a rewrite. Or, if you pushed it to a new branch, you'll have to edit your PR to point to the new branch.

tomballgithub:
git remote add upstream https://github.com/PokeAlarm/PokeAlarm.git
git fetch upstream
git -i rebase upstream/dev
git push -f origin SponsoredRaids

Tried again and then reset it back the way it was.  I get 41 files changed and most of it has nothing to do with my PR.  I saw the comment about doing an -i rebase, which I did, but not of the commits listed needed to be commented out, they were ALL related to my PR.

tomballgithub:
I believe 'origin' is the master branch

tomballgithub:
All rebased correctly.

FYI, I had someone ask me coincidentally if there was a way they could make a manager that only triggered on raids with the word 'park' in them.  This method will support that as you guys recommended.

tomballgithub:
And, the way this is written, with the filter being optional, no existing filter files need to be changed.  So although I have 100+ sets of manager files only 1 is for sponsored raids and so that is all that needs editing

kvangent:
@tomballgithub the travisci check is showing some pep8 coding style violations. Could you please fix those so we can pull this?

You can see them by clicking 'Details' next to the travis status check.

You can run flake8 on your own machine by installing via pip, and then running `flake8 --show-statistics --show-source`

tomballgithub:
All fixed...

PokeAlarm/Manager.py
```diff
@@ -981,8 +981,11 @@ def process_egg(self, egg):
 
         # Check if egg gym filter has a contains field and if so check it
         if len(self.__egg_settings['contains']) > 0:
-            if not any(x in gym_info['name'].lower() for x in self.__egg_settings['contains']):
-                log.info("Egg {} ignored: gym name did not match contains filter.".format(gym_info['name'].lower()))
+            if not any(x in gym_info['name'].lower()
+                       for x in self.__egg_settings['contains']):
```
kvangent: Have you tested this? It looks like your indentation might be off, and only the first gym_name contains will be checked

tomballgithub:
I tested and verified that the checking of the gym name with the formatting change works as desired

I did find that spreading the log.info format string across multiple lines with + did not work and all {} were printed as {} with no substitution.  Removing the + and letting python automatically concatenate adjacent strings is acceptable and works.

Changed to gym_id in the log message instead of the gym name for consistency

tomballgithub:
Someone who pulled my fork/branch directly says this is not triggering as expected.  I will investigate including adding log.debug information for future

blakewenloe:
It would be amazing if I could check against an array of gym names that I can input/change as needed. That would only require one filter file.

tomballgithub:
I have extensive logs showing that this feature works.  I added some optional debug logging.  And fixed Travis.

kvangent:
Can you confirm this is up to date with the dev branch? It says it's out of date and failing Tavis.

tomballgithub:
Yes, there was a travis problem of two new debug lines being too long.  Fixed and passing...

tomballgithub:
And I confirmed the person who was having issues with my branch is using Monocle and was not receiving the gym_name information via Monocle which is why they were having issues

filters.json.example
billyjbryant: Filter examples need to be updated with the change in verbiage and to match the actual checks.

filters.json.example
```diff
@@ -426,11 +426,13 @@
...
         "min_level": "1",
         "max_level": "5"
     },
     "raids":{
         "enabled":"True",
+        "gymname_contains":["sprint","level badge","phone skin","starbucks"],
```
billyjbryant: Same as above.

filters.json.example
tomballgithub: 'gymname_contains' is the verbiage.  can you clarify what you are asking?

filters.json.example
tomballgithub: 'gymname_contains' is the verbiage.  can you clarify what you are asking?

billyjbryant: Nevermind, I misread. You are good.

PR #502, opened 2017-11-24
Event Rework

kvangent:
This PR is the start of how Events are handled in PA. A new class called 'Event' with several sub-classes have been added to represent information coming in via webhooks. Currently none of them are implemented, but they will be used in later work. 

We'll be skipping the wiki update for now since this currently has no impact on the code base.

### DIFF
--- a/PokeAlarm/Events/Gym.py
+++ b/PokeAlarm/Events/Gym.py
@@ -1,7 +1,7 @@
 # Standard Library Imports
 # 3rd Party Imports
 # Local Imports
-from PokeAlarm.Utils import get_gmaps_link, get_applemaps_link, get_time_as_str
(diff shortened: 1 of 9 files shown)


PokeAlarm/Events/Gym.py
```diff
@@ -0,0 +1,66 @@
...
+            'old_team_leader': locale.get_leader_name(self.old_team_id),
+            'new_team': locale.get_team_name(self.new_team_id),
+            'new_team_id': self.new_team_id,
+            'new_team_leader': locale.get_leader_name(self.new_team_id),
+
+            # Details
+            'gym_name': self.gym_name,
+            'gym_description': self.gym_description,
+            'gym_image_url': self.gym_image_url,
+        }
```
billyjbryant: This doesn't account for the changes I've made to gyms in the WebhookStructs

PR #504, opened 2017-12-04
Webhook gym from cache

mali055:
## Description
It add new functionality to the webhook_tester, by letting it read a user specified cache file, and verify the gym id entered by the user to extract the gym_name from the cache, then send the webhook with a valid entry.

## Type of Change

## Motivation and Context
Was annoyed that gym tester was not helping anyone test gym webhooks properly, this will help remedy some of the testing problems.

## How Has This Been Tested?
Tested myself, reads cache file, relative to webhook_tester file, and defaults to manager_0, and exits in the case neither are present. Gets name from cache, and sets the id and name in the json.

## Wiki Update

## Checklist


tools/webhook_test.py
```diff
@@ -258,6 +317,10 @@ def get_and_validate_team():
     else:
         print "Egg level invalid. Assuming level 5"
 elif type == whtypes["5"]:
+    print "Enter cache file name to verify the gym (default: manager_0)\n>",
+    cache_or_invalid()
```
LegitDongo: This forces the user to use file caching, while this is an optional feature.
You aren't able to send a raid webhook without file caching, so it's better to have this as optional as well.

tools/webhook_test.py
```diff
@@ -125,6 +145,46 @@ def int_or_default(input_parm):
         raw_input(), payload["message"][input_parm])
 
 
+def get_gym_info(gym_id):
+    """ Gets the information about the gym. """
+    return _gym_info.get(gym_id, _default_gym_info)
+
+
+def check_gym(questionable_input):
+    if get_gym_info(questionable_input)['name'] != "unknown":
```
LegitDongo: Nobody will be able to get the gym id without jumping through a bunch of hoops.
This would probably be better as presenting as a list gym_name options and allowing the user to pick one to make it easier on the user.
I'd imagine nobody wants to remember a gym id :smile:

tools/webhook_test.py
```diff
@@ -125,6 +145,46 @@ def int_or_default(input_parm):
...
+
+
+def gym_or_invalid(prm, prm2):
+    payload["message"][prm], payload["message"][prm2] = check_gym(raw_input())
+
+
+def cache_or_invalid():
+    path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    input = raw_input()
+    if os.path.exists(os.path.join(path, "cache", "{}.cache".format(input))):
```
LegitDongo: It would probably also be better to present the managers found in a list presented to the user and allowing them to pick one.

tools/webhook_test.py
```diff
@@ -125,6 +145,46 @@ def int_or_default(input_parm):
...
+    """ Gets the information about the gym. """
+    return _gym_info.get(gym_id, _default_gym_info)
+
+
+def check_gym(questionable_input):
+    if get_gym_info(questionable_input)['name'] != "unknown":
+        print "Gym found = {}".format(get_gym_info(questionable_input)['name'])
+        return questionable_input, get_gym_info(questionable_input)['name']
+    else:
+        print "Not a valid gym. Please try again.."
```
LegitDongo: Should follow the standards with the rest of the program of the newline and the `>` since this could happen a lot 😄 
Ex. `print "Not a valid gym. Please try again..\n>",

tools/webhook_test.py
mali055: This won't be practical if there are alot of gyms in the users cache,

tools/webhook_test.py
```diff
@@ -125,6 +147,76 @@ def int_or_default(input_parm):
...
+    if get_gym_info(questionable_input)['name'] != "unknown":
+        print "Gym found = {}".format(get_gym_info(questionable_input)['name'])
+        return questionable_input, get_gym_info(questionable_input)['name']
+    else:
+        print "Not a valid gym. Please try again..\n>"
+        check_gym(raw_input())
+
+
+def gym_or_invalid(prm, prm2):
+    payload["message"][prm], payload["message"][prm2] = check_gym(raw_input())
```
LegitDongo: When I run this and get to the `Enter gym id for raid` prompt, and I put in an empty string once, then a valid gym id the next time, this errors out.
It appears this is because of the recursive calls of `check_gym`. It looks like that it returns the data accurately once, and gives invalid data the other times.
```
Gym found = The Battle of Carthage
  File "C:\Program Files\JetBrains\PyCharm 2017.2.4\helpers\pydev\pydevd.py", line 1599, in <module>
    globals = debugger.run(setup['file'], None, None, is_module)
  File "C:\Program Files\JetBrains\PyCharm 2017.2.4\helpers\pydev\pydevd.py", line 1026, in run
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "C:/PycharmProjects/mali/tools/webhook_test.py", line 444, in <module>
    gym_or_invalid("gym_id", "gym_name")
  File "C:/PycharmProjects/mali/tools/webhook_test.py", line 238, in gym_or_invalid
    payload["message"][prm], payload["message"][prm2] = check_gym(raw_input())
TypeError: 'NoneType' object is not iterable
```

tools/webhook_test.py
```diff
@@ -125,6 +147,76 @@ def int_or_default(input_parm):
...
+    """ Gets the information about the gym. """
+    return _gym_info.get(gym_id, _default_gym_info)
+
+
+def check_gym(questionable_input):
+    if get_gym_info(questionable_input)['name'] != "unknown":
+        print "Gym found = {}".format(get_gym_info(questionable_input)['name'])
+        return questionable_input, get_gym_info(questionable_input)['name']
+    else:
+        print "Not a valid gym. Please try again..\n>"
```
LegitDongo: Forgot the `,` at the end :grin:

PR #519, opened 2017-12-17
New Filters Rework

kvangent:
This is the implementation of the new filters rework. Filter's have changed significantly. There are now the following 5 sections. Each has `enabled`, `defaults`, and `filters` subsections. `enable` can be either true or false. `defaults` are parameters that are automatically applied to all filters, unless overrode in that specific filter. `filters` is a list of name -> filters that decided which events get notified.

It's also worth noting that `missing_info` is fundamentally different from `ignore_missing` - when a filter is checking an event, it will skip any parameters that are unknown. If iv info is missing and `min_iv` is set, `missing_info` will allow it if `true` and reject if `false`.

Geofences have also changed - they are no longer checked for every filter. Instead each filter has a `geofences` section in which a list of geofence names can be listed. Filters will only check the geofences they have been assigned. 

Another new change is `customer_dts`. With this parameter, you can add a map of key-value pairs that act as custom DTS for that filter. With `"custom_dts": { "key1": "value1", "key2": "value2" }`, `<key1>` would be replaced as `value1`. 

The is a significant change that will break older configs, so don't try it out if you don't have time. There are numerous DTS changes as well, so it's recommended you use the Webhook Tester to ensure things are working as intended. 

**!!! Warning !!!**: This feature is still new, so make sure to keep back ups of your older configs in case you run into problems. 

### Monsters

**filters**
```json
"monsters":{
        "enabled": true,
        "defaults": {
        },
        "filters": {
            "filter-name": {
                "monsters": [ "Bulbasaur", 2, 3 ],
                "min_dist": 0, "max_dist": "inf",
                "min_lvl": 0, "max_lvl": 40, "min_atk": 0, "max_atk": 15,
                "min_def": 0, "max_def": 15, "min_sta": 0, "max_sta": 15,
                "min_iv": 0.0, "max_iv": 100,
                "form_ids": [ 0 ],
                "quick_moves": [ "Vine Whip", "Tackle"],
                "charge_moves": [ "Sludge Bomb", "Seed Bomb"],
                "genders": [ "male", "female", "neutral"],
                "min_height": 0, "max_height": "inf",
                "min_weight": 0, "max_weight": "inf",
                "sizes": [ "tiny", "small", "normal", "large", "big" ],
                "geofences": [ "Central Park" ],
                "custom_dts": { "key1": "value1", "key2": "value2" },
                "missing_info": false
            }
        }
    },
```

**dts**
```
[...]

PR #520, opened 2017-12-18
Added ability to override GymName by providing a list of gyms with ID…

blsmit5728:
Added ability to override GymName by providing a list of gyms with ID's,name,lat,lon in order to fill cache on startup. Also added a role_pokemon filter which allows the user to specify certain pokemon that will cause discord to tage a role in the server upon posting the webhook. Included also is a python tool to get the base16 ID's of the gyms from the RocketMap Database, convert them to base64 for use with PA and create an output similar to the gyms.csv file needed.

## Description
This PR will add the ability to override the in-game Gym names to make it where a user can use gym names that may be easier in identifying which gym a raid or egg is at. I have also created a filter that allows the user to specify pokemon that will trigger a role tag in discord. 

## Type of Change

## Motivation and Context
We were tired of not being able to decipher between which Starbucks without having to "name" them in the chat. Also, there were certain gyms that have very strange names in the game, but the slang we used was much easier to remember. 

## How Has This Been Tested?
I currently run this in my production setup, our 
image
## Wiki Update

- [x ] This change requires an update to the Wiki.

## Checklist
- [x ] This change only changes what is necessary to the feature.


kvangent:
This will need to be rebased onto #519, as it will be pulled shortly and involves significant reqorks. It also has numerous PEP8 code style violations (which is why it is failing in Travis.)

Additionally, it's best to only submit one feature at a time when submitting a PR  - this makes it easier to review and accept/reject certain features without blocking them all at once.

Finally, this feature (ie the GymName override) would probably be best implemented as a custom gym cache object based on the existing interface. It could load a file on initialization and prioritize ids given to it over what is cached. This would also keep it more contained and easier to maintain.

blsmit5728:
Thanks @kvangent I'll rebase this onto #519 once it's pulled and work on building it into a custom Gym Cache object. Also, I'll remove the extra gym_info I added since it's not really part of this feature. Lastly, I'll work on the PEP8 stuff. Thanks for the feedback.

kvangent:
Closing this since it is fairly diverged - Please feel free to reopen when the feedback has been addressed.

PR #537, opened 2018-01-07
Add weather to mon and raid events

evilmoses:
## Description
Pokemon Go now boosts certain Pokemon spawns and raid bosses when the associated type matches the correct weather. 

As a starter, let me say that Python is still very new to me and I imagine you guys will tear this code apart, but it may be useful as a starting point. I use Monocle, so cannot say whether RM is yet dealing with weather or not, but this PR presume that weather is sent to the webhook. I also have not tested this with Patch 3.5 as I am still in the process of merging, but the concept should remain valid here.

__Protobuf info__

For Pokemon, the weather information is found within weather_boosted_condition and this can be sent to the webhook very easily.

For raids, the weather information is only found in the lobby section of the raid, and therefore scanning software cannot pick it up. The fork of Monocle I use gets the weather by working out which s2 cell the raid is in, and getting the current weather for that s2 cell. This weather condition is then sent to the webhook.

If the weather is sent to the webhook with the name 'weather', this PR will output the following:

__Pokemon__

Will output the weather boosted condition in dts <weather>

__Raids__

Will output the weather boosted condition in dts <weather> and will also output the correct <min_cp> and <max_cp> using boss level 25 instead of 20.

## Type of Change

## How Has This Been Tested?

This code has been running on my setup for weeks. I have 13 scan instances and around 35 PA managers. I've had no issues with it. As mentioned above, it has not been tested on 3.5, but I though I would submit the PR to start the ball rolling.

mon_weather

raid_weather

## Checklist


PokeAlarm/Utils.py
```diff
@@ -273,6 +273,45 @@ def get_pokemon_gender(gender):
...
+        with open(file_, 'r') as f:
+            j = json.loads(f.read())
+        for id_ in j:
+            get_base_types.info[int(id_)] = {
+                "type1": j[id_].get('type1'),
+                "type2": j[id_].get('type2')
+            }
+    return get_base_types.info.get(pokemon_id)
+    
+def is_raid_boss_weather_boosted(pokemon_id, weather_id):
```
kvangent: This function could use some optimization.

Ideally we would just keep a dict of weather_id -> boosted_types and returning something like `type1 in dict or type2 in dict`. 

We should also try to strive for using types as `ints` (much faster to compare and less troubles with encoding) (I know that base_stats doesn't have this currently, but we're working on it)

MARTINI1:
3 DTS could be created by combining this PR with #540

weather_id, weather_name and weather_emoji

same as this https://github.com/PokeAlarm/PokeAlarm/commit/0192258ba419d2d4aa7f7898e59a9f86ce1deb24

and like the other PR, it's a good idea to add support for locale 👍

kvangent:
@evilmoses I think `weather_id` and `weather_emoji` would be reasonable additions. Do you mind adding them in?

evilmoses:
Added the two new DTS and tested.

kvangent:
I think this looks really solid, I'm only concerned about the emojis working with other services.

LegitDongo:
We all discussed how the defaults and naming for scheme for weather should be, and here's what we came up with:
For defaults it'd be good to have `weather_emoji` default to an empty string.
Have `weather` default to `None` like you have it, and add an extra DTS for `weather_or_none`, in case someone has a use-case for it.

evilmoses:
Added weather_or_empty
Set weather_emoji to '' when not found
Set weather_id to 0 when not found

nhumber:
emoj not working with discord, to get it working i had to modify utils.py
added this at line 1

`# -*- coding: utf-8 -*-`

and changed your weather emoj code in there to this:

```
def get_weather_emoji(weather_id):
    return {
            1: u"☀️",
            2: u"☔️",
            3: u"⛅",
            4: u"☁️",
            5: u"💨",
            6: u"⛄️",
            7: u"🌁"
    }.get(weather_id, '')

```

nhumber:
another issue is when using RM (sloppys for at least) this PR breaks raid alerts.

2018-01-16 07:02:05,422 [      manager_0][   Manager][   ERROR] Encountered error during processing: KeyError: '0'

evilmoses:
@nhumber done :)

nhumber:
latest commit fixed raid alert error with RM, emoji DTS is also working in username with discord.

LegitDongo:
From what I've tested already this looks pretty solid.
All it needs is a rebase off the dev branch and another quick test to make sure nothing :b:roke, then we good to go 👍

kvangent:
Ok so we changed a couple of things to be consistent with the other DTS. `weather_id` defaults to `unknown` now so that users can tell when the weather was just not sent or when there is no weather. 

`weather` will be None if weather_id is 0
`weather_or_empty` and `weather_emoji`

@LegitDongo also added filtering and unittests

PR #545, opened 2018-01-10
Fix cardinal directions

crhbetz:
Noticed that `<direction>` DTS always resulted in a `?`. Found `get_cardinal_dir` in utils and references in the *Filter.py files that cardinal directions should be handled by Manager. However, it didn't even import `get_cardinal_dir`. It was all there pre-3.5, so I tried to restore it.

I'm still a beginner in Python, so I'd be glad if you could point out any and all mistakes I made in creating this PR. Thanks :)

## Description
Restoring functionality to the `<direction>` DTS 

## Type of Change

## Motivation and Context

bug fix
## How Has This Been Tested?

Tested on a test instance with Pokemon events from webhook_test.py
## Wiki Update

## Checklist

### DIFF
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -22,7 +22,7 @@ from Geofence import load_geofence_file
 from Locale import Locale
 from LocationServices import location_service_factory
 from Utils import (get_earth_dist, get_path, require_and_remove_key,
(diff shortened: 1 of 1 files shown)


kvangent:
Look good. Thanks for taking the time to do this.

PR #546, opened 2018-01-10
 Gym_Name not working for raid notifications.

Markair:
## Description
Adding the ", data.get('id')" fixes the issue and allows the user to post gym_name successfully.

## Type of Change

## Motivation and Context
Allows users to post gym name on Egg and Raid notifications.

## How Has This Been Tested?
Tested locally through Webhook_test and in through local Alarm.

## Wiki Update

## Checklist

### DIFF
--- a/PokeAlarm/Events/RaidEvent.py
+++ b/PokeAlarm/Events/RaidEvent.py
@@ -18,7 +18,7 @@ class RaidEvent(BaseEvent):
         check_for_none = BaseEvent.check_for_none
 
         # Identification
-        self.gym_id = data.get('gym_id')
+        self.gym_id = data.get('gym_id', data.get('id'))
 
         # Time Remaining
         self.raid_end = datetime.utcfromtimestamp(


kvangent:
Is this for Monocle? The webhook tester uses 'gym_id' and RM sends it as 'gym_id', so I guess I'm sure why this is necessary (or why testing with the webhook tester is useful here).

Markair:
Using RM data.
The issue I was seeing is that it wasn't posting the gym_name to either the egg or raid notifications, it was always posting unknown, even after refreshing my cache.
I ran tests through the webhook_test and for gym notifications, it was showing me a list from my cache and notifying on the gym name for the gym notifications. 
When I did the same thing for egg or raid tests, it wouldn't load the cache gym list or send the gym name in the notification.
Once I'd added , data.get('id') to egg/raid event, it showed the gym list from the cache and notified the gym name correctly and not unknown in the notifications.

kvangent:
> data.get('id') to egg/raid event
Currently only raid is changed in this PR. Is that intentional? 

As far as I can tell, RM, Monocle, and the webhook tester all send it as `gym_id`. I've confirmed that gym_names cache correctly for me (both raids and eggs) without this change.  I'm fine with it if it is really needed, but I'd like to confirm the reason we need to make this adjustment.

Markair:
I'm new to using Git, I didn't k ow how to check both in on the same PR (still learning.) 

I'll add a screenshot later of the differences I see when the change isn't included.

Markair:
When testing using caching to file, all gym names were shown as for eggs and raids:
screenshot_2018-01-13-09-38-25

For gym notifications it was working as follows:
screenshot_2018-01-13-09-38-39

After updating the raidevent.py/eggevent.py it notifies as follows:
screenshot_2018-01-13-09-38-53

kvangent:
I'm assuming this was closed because your problem was something else?

@Markair To follow up on git, you can stage multiple files before you commit (`git add <filename>` or `git add -A` for all changed files), which is how you would change multiple things at once. You can also put multiple commits onto a single branch.

PR #551, opened 2018-01-14
Added functionality to set desired pokemons and raids on filters file by sending private messages

hubertokf:
## Description
This commit adds the functionality to ask to telegram by private messages to include or exclude pokemons from raid or monsters filters file.

In extension to this commit, requirements.txt must be updated to include telepot==12.5.

Future modification is to reload filters on execution.

## Type of Change

## Motivation and Context
It make easy to select what pokemons must be alerted without the need to edit a json file.

## How Has This Been Tested?
It was tested on 3 simultaneous instances.

## Wiki Update

## Checklist

### DIFF
--- a/PokeAlarm/Alarms/Telegram/TelegramAlarm.py
+++ b/PokeAlarm/Alarms/Telegram/TelegramAlarm.py
@@ -7,6 +7,7 @@ import telepot
 # Local Imports
 from PokeAlarm.Alarms import Alarm
 from Stickers import sticker_list
+from TelegramConfiguratorBot import ConfiguratorBot
 from PokeAlarm.Utils import parse_boolean, require_and_remove_key, \
     reject_leftover_parameters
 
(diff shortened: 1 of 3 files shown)


hubertokf:
Is it necessary to follow max line length?

kvangent:
This seems to have a bunch of issues.

* It'll only change the filter named "filter-name". Other filters won't be changed. 
* There isn't any input verification, so incorrectly spell monsters will crash the whole thing when it next restarts.
* Changes done won't take affect until PA is restarted, but there isn't any way to do that, making this whole change not very useful. 

I'm going to close this for now because I don't want people to use it until it is a little more robust.

Ideally, it would you would be able to do everything the filters file can do. ei `add filter monsters "new-filter":{ "monsters": [ 1, 2, 3 ]}` or `remove filter monsters "new-filter"`. However it also needs to be robust enough that incorrect validation won't fail.

kvangent:
Also, yes, you need to follow the pep8 style checks.

hubertokf:
ok, will enhance and then make a new commit. Thanks

PR #554, opened 2018-01-15
Enable custom_dts for Telegram chat_id

crhbetz:
With this simple change you can use a custom_dts to set Telegram's chat_id, like it is already possible with Discord Webhook or Slack Channel, according to wiki.

## Description
Use Alarms.replace on Telegram chat_id

## Type of Change

Not sure if leaving this out was an oversight or intentional.

## Motivation and Context
Being able to set chat_id through custom_dts like with Discord Webhook or Slack Channel enables you to significantly reduce your manager count if you're sending messages to different chats. Without this change, you need one manager per chat_id. I can't see a reason not to do this specifically for Telegram chat_id when it's possible for other alarms.

## How Has This Been Tested?
Successfully used the change on my production installation, serving over 50 individual user chats over Telegram with individual Pokemon Alarms, for 5 days.

## Wiki Update

## Checklist

### DIFF
--- a/PokeAlarm/Alarms/Telegram/TelegramAlarm.py
+++ b/PokeAlarm/Alarms/Telegram/TelegramAlarm.py
@@ -116,14 +116,14 @@ class TelegramAlarm(Alarm):
     # Send Alert to Telegram
     def send_alert(self, alert, info, sticker_id=None):
         if sticker_id:
-            self.send_sticker(alert['chat_id'], sticker_id)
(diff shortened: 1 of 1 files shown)


kvangent:
Telegram refresh in #556 also includes this change.

crhbetz:
Can confirm it's working on #556 , exactly like with this. No need for this PR, then, so I'll close it.

PR #555, opened 2018-01-15
Manager: Do not stop checking filters after first match

crhbetz:
## Description
Applying this PR will enable PokeAlarm to match multiple filters on a single event. 

I discussed it with @billyjbryant already who explained that this change breaks the FIFO idea for filters, which was introduced to reduce memory footprint. It will also change behavior with existing configurations: Users might have set up their filters with the idea in mind that only the first match will be processed, leading to unexpected additional messages after this change. However, he encouraged me to submit the PR for discussion.

## Type of Change

## Motivation and Context
I'm offering individualized Pokemon alarms on a per-user basis (through Telegram). Considering every single user might want to be informed about a wild Tyranitar, I'm forced to use a dedicated manager for every user, or only the lucky one to be the first in the filters file would be notified. At 50 users, I saw a total memory footprint of 650MB.

After applying this change, I can process all individual users through one single manager. The memory footprint now settles at 55 to 60MB. For me, the ability to reduce from 50 to 1 manager significantly outweighs the memory footprint reduction from application of the FIFO principle.

Another example would be raid channels with overlapping or layered areas. Let's say a discord server would have separate "New York" and "Lower Manhattan" channels. A raid in Lower Manhattan should  show up in both of these channels, so it would need to continue checking further filters after matching against the first one.

## How Has This Been Tested?
I've been using this along with https://github.com/PokeAlarm/PokeAlarm/pull/554 on my production environment serving 50+ individual users and 5 public channels and did not encounter any issues so far.

## Wiki Update

## Checklist

### DIFF
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -485,31 +485,31 @@ class Manager(object):
         passed = False
         for name, f in self.__mon_filters.iteritems():
             passed = f.check_event(mon) and self.check_geofences(f, mon)
-            if passed:  # Stop checking
+            if passed:
                 mon.custom_dts = f.custom_dts
-                break
+
+                # Generate the DTS for the event
+                dts = mon.generate_dts(self.__locale)
+
+                if self.__loc_service:
+                    self.__loc_service.add_optional_arguments(
+                        self.__location, [mon.lat, mon.lng], dts)
+
+                if self.__quiet is False:
+                    log.info("{} monster notification has been triggered!"
+                        .format(mon.name))
+
+                threads = []
+                # Spawn notifications in threads so they can work in background
+                for alarm in self.__alarms:
+                    threads.append(gevent.spawn(alarm.pokemon_alert, dts))
+                gevent.sleep(0)  # explict context yield
+
+                for thread in threads:
+                    thread.join()
+
         if not passed:  # Monster was rejected by all filters
             return
-
-        # Generate the DTS for the event
-        dts = mon.generate_dts(self.__locale)
-
-        if self.__loc_service:
-            self.__loc_service.add_optional_arguments(
-                self.__location, [mon.lat, mon.lng], dts)
-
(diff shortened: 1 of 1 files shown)


sebastienvercammen:
I have something related coming up that'll change the discussion on memory footprint when using multiple managers. It's not finished yet, and I have other work to take care of first, but keep it in mind. As always, Deadly, Dongo and the rest can poke me on Discord if they have more immediate questions.

kvangent:
So the change in filters wasn't really targeted at a reduction in overall memory usage, but instead more so on reducing the number of managers being used. Manager's were never really intended to be run by the dozens (or hundreds), but that is obviously the point at where we are at today. This is a problem because Managers compete with the original web-server process for CPU cycles. Someone can correct me if I'm wrong, but I don't think that memory footprint is really a problem at the moment (not to say it shouldn't be observed and minimized, but that it isn't causing critical bottlenecks).

Anyway, the decision for making filters be 'one and done' was intentional and for a couple of reasons. First, it was closest to how PA previously handled 'multi-filtering'. If any Event passes any Filter, it gets passed to the Alarms to trigger notifications. This seemed to be the more obvious behavior since it was consistent with previous behavior (although now I'm not sure this was the case). Secondly, it was the more efficient way to treat them. Checking less filters means more work, which improves overall event throughput. Finally, the biggest reason was that the use case like yours could still be accomplished with multiple managers, but the inverse use case (getting only one notification even when matching multiple filters) could not. As such, we decided to go with the default that would work for everyone (even if it's a little less work for certain setups)

In v3.6 we are planning to introduce a feature called 'Rules'. I originally intended to include this feature in v3.5, but I didn't get it done in time and didn't want to hold up that patch any longer. The TL;DR of this is that you will be able to choose which filters (and what order) are checked and which alarms they trigger. 

In the short term, I'm fine with leaving this PR open so that users who would like this functionality now can use it.

PR #556, opened 2018-01-16
Telegram Refresh

kvangent:
This is a rework/rewrite of the telegram alarm. It's had a few issues for a while and the API has changed significantly since the original was wrote.

There are still some issues (namely the default images aren't very high quality), but I'm opening the PR to ask for help with testing.

If you have any issues with this, please comment on the issue below. 

## Changes

* Telegram now uses markdown instead of html. This was done to reduce conflicts with incorrect DTS causing the entire message to be rejected by the service.
* 'title' and 'body' -> `message`: merged into a single 'message' to better represent how the Telegram API actually treats messages. Old behavior can be mimiced with: `*TITLE GOES HERE*\n BODY GOES HERE`
* Alert titles have been updated to reflect new `filters.json` sections: `monsters`, `stops`, `gyms`, `eggs`, and `raids`
* A bunch of fields have changed in order to be more understandable. All of these fields are configurable at the alert level, and default to the Alarm level when not specified:
  * `bot_token`: Can now be set at an Alert level. DTS compatible.
  * `chat_id`: Now DTS compatible.
  * `stickers` -> `sticker`: `true` for sticker with message, `false` for no sticker. default to `true`.
  * `sticker_notify`: Whether or not sticker messages causes a notification. `true` or `false`
  * `sticker_url`: Url to be used for the sticker. Must be .webp file. DTS compatible.
  * `location`-> `map`: `true` for map after message, `false` for no map. default to `true`.
  * `map_notify`:  Whether or not map messages causes a notification. `true` or `false`
  * `venue`: Sends the map and message in a single condensed format. `true` or `false`
  * `venue_notify`: Whether or not venue messages causes a notification. `true` or `false`
  * `max_retries`: Max attempts to send for each message. (Telegram no longer uses the command line equivalent)


PokeAlarm/Alarms/Telegram/TelegramAlarm.py
```diff
@@ -23,194 +23,259 @@
...
         }
-        try_sending(log, self.connect, "Telegram (venue)",
-                    self.__client.sendVenue, args)
+        try_sending(
+            log, self.connect, "Telegram (LOC)", self.send_webhook, args,
+            max_attempts)
 
-    # Send a location message to telegram
-    def send_location(self, alert, info):
+    def send_venue(self, token, chat_id, lat, lng, max_attempts):
```
crhbetz: The new send_venue lacks the 'message' text. You're only passing/using chat_id and the location. This will send a venue without any text.

kvangent: Yeah that is also something I need to add. My intent is to split it on the first new line.

kvangent:
@crhbetz I've updated venue to work correctly. Please let me know if you see any other issues.

crhbetz:
I played around with it a bit more and I'm planning to try it on my production instances tomorrow. Local testing looking great so far, no issues with the basics.

One detail: I don't really get how `map_notify` and the other similar options should work. If I set these to False, I correctly see `'disable_notification': True,` in the DEBUG log:
`2018-01-16 22:34:13,518 [       channels][  Telegram][   DEBUG] {'latitude': xx.xxx, 'disable_notification': True, 'chat_id': '-100zzz', 'longitude': yy.yyy}`
Still when sending sticker, text and location with notfications for sticker and location disabled, I see "3 new messages" on my phone.

kvangent:
@crhbetz I also get confused by it (the naming isn't very great if you ask me). Telegram's API describes the 'disable_notification' as the following: 

> Sends the message silently. Users will receive a notification with no sound.

In the past there were issues with this working differently on Android and iOS phones, and I'm not sure if Telegram has become more consistent in this area or not.

scanom:
I'm trying to test the PR but I cannot get it to start the instance:
`pickle.PicklingError: Can't pickle <class 'PokeAlarm.Alarms.Telegram.TelegramAlarm.Alert'>: it's not found as PokeAlarm.Alarms.Telegram.TelegramAlarm.Alert`

crhbetz:
@kvangent Ah, I didn't expect this to be the intended behaviour...
`'disable_notification': True` -> `Users will receive a notification` - great naming indeed. It's working as intended then.

PokeAlarm/Alarms/Telegram/TelegramAlarm.py
```diff
@@ -23,194 +23,263 @@
...
         args = {
-            'chat_id': chat_id,
-            'sticker': unicode(sticker_id),
-            'disable_notification': 'True'
+            'url': "https://api.telegram.org/bot{}/sendMessage".format(token),
+            'payload': {
+                'chat_id': chat_id,
+                'text': message,
+                'parse_mode': 'Markdown',
+                'disable_web_page_preview': False,
```
MARTINI1: change this to 'True' please

crhbetz:
I'm getting duplicate messages occasionally (1 in 100 or so), with this in log:
`2018-01-17 13:42:51,492 [       channels][  Telegram][   ERROR] Encountered error while sending notification (ReadTimeout: HTTPSConnectionPool(host='api.telegram.org', port=443): Read timed out. (read timeout=5))`                                                                                                    
`2018-01-17 13:42:51,493 [       channels][  Telegram][ WARNING] Telegram (LOC) is having connection issues. 1 attempt of 3.`
I haven't seen duplicate messages before applying this PR. Is the read timeout too low maybe? I can't figure out what it was before.

PokeAlarm/Alarms/Telegram/TelegramAlarm.py
kvangent: done

kvangent:
@crhbetz I've moved the timeout to 30 seconds to match telepot's (which seems like overkill), but we'll start there and see if that fixes the problem.

@scanom I've switched to a name tuple which I'm hoping will play better with windows pickling.

scanom:
@kvangent No luck either with named tuple, exactly the same error:
`pickle.PicklingError: Can't pickle <class 'PokeAlarm.Alarms.Telegram.TelegramAlarm.Alert'>: it's not found as PokeAlarm.Alarms.Telegram.TelegramAlarm.Alert`

PokeAlarm/Alarms/Telegram/TelegramAlarm.py
```diff
@@ -23,194 +24,270 @@
 
 
 class TelegramAlarm(Alarm):
-    _defaults = {
-        'pokemon': {
-            # 'chat_id': If no default, required
-            'title': "A wild <mon_name> has appeared!",
-            'body': "Available until <24h_time> (<time_left>)."
+
+    Alert = namedtuple(
```
scanom: Leave namedtuple outside of the class.

PokeAlarm/Alarms/Telegram/TelegramAlarm.py
```diff
@@ -23,194 +24,270 @@
...
-        self.__client = telepot.Bot(self.__bot_token)
+        pass
+
+    # Set the appropriate settings for each alert
+    def create_alert_settings(self, kind, settings, alert_defaults):
+        default = TelegramAlarm._defaults[kind]
+        default.update(alert_defaults)
+        settings = Alarm.pop_type(settings, kind, dict, {})
+
+        alert = TelegramAlarm.Alert(
```
scanom: alert = Alert(

scanom: So far so good, working well with the changes suggested above. Anyway the images quality is not the same as old stickers,, you may let the users pick their own webp URLs or change the images.

kvangent:
@scanom I've rebased on dev which no longer uses multiprocessing (so shouldn't have any pickle issues with windows).

scanom:
@kvangent working perfect after rebase ;)

PR #558, opened 2018-01-18
Performance fixes. 

kvangent:
Remove multiprocessing and gipc and substitute greenlets instead. The queues in multiprocessing/gipc are process blocking which is causing unnecessary overhead from context switches. 

Thanks to @sebastienvercammen for pointing out the flaw in our setup.

### DIFF
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -331,9 +331,6 @@ class Manager(object):
     def setup_in_process(self):
 
         # Update config
-        config['TIMEZONE'] = self.__timezone
-        config['API_KEY'] = self.__google_key
-        config['UNITS'] = self.__units
         config['DEBUG'] = self.__debug
         config['ROOT_PATH'] = os.path.abspath(
             "{}/..".format(os.path.dirname(__file__)))
@@ -487,7 +484,7 @@ class Manager(object):
             return
 
         # Generate the DTS for the event
-        dts = mon.generate_dts(self.__locale)
+        dts = mon.generate_dts(self.__locale, self.__timezone)
 
         if self.__loc_service:
             self.__loc_service.add_optional_arguments(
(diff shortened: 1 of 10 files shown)


PokeAlarm/Manager.py
```diff
@@ -110,11 +109,11 @@ def stop(self):
...
-        if self.__process.is_alive():
+        self.__process.join(timeout=20)
+        if not self.__process.ready():
             log.warning("Manager {} could not be stopped in time!"
-                        + " Forcing process to stop.")
-            self.__process.terminate()
+                        " Forcing process to stop.".format(self.__name))
+            self.__process.kill(timeout=2, block=True)  # Force stop
         else:
             log.info("Manager {} successfully stopped!".format(self.__name))
```
billyjbryant: In normal closing, it does not appear that the system gets to this point before the process exits.

jgehrcke:
>  The queues in multiprocessing/gipc are process blocking

gipc author here. gipc's inter-process communication is not blocking gevent's machinery. That's the entire point of gipc primitives :)

But yes, most of `multiprocessing` is not blocking cooperatively.

So, feel free to use gipc in your gevent-driven code!

PokeAlarm/Manager.py
```diff
@@ -327,18 +326,11 @@ def set_optional_args(self, line):
 
     # Start it up
     def start(self):
-        self.__process = gipc.start_process(
-            target=self.run, args=(), name=self.__name)
+        self.__process = gevent.spawn(self.run)
```
jgehrcke, line 329: If you ever need to off-load this work into a child process again then the way you did that before was the correct one. However, for doing inter-process communication with the child in a way so that the gevent machinery in the parent is not negatively affected the `gipc.pipe()` is supposed to be used.

kvangent: Could you expand on what you mean here? Are you saying the problem we observed with `gevent.Queue()` and that `gipc.Pipe()` would have prevented the problem? 

It's obviously been a few months, but I vaguely recall testing `gipc.Pipe()`, `gevent.Queue()`, and `gipc.Pipe()` and still seeing significant performance drops compared to the single process. I thought I had found something in the gipc documentation to support that statement, but perhaps I'm mis-remembering.

PR #559, opened 2018-01-19
Change "gym_name_contains" to ignore case. 

kvangent:
This will change "gym_name_contains" to ignore case, making it easier for users to filter on this name.


kvangent:
Added casting to unicode strings to prevent unintentional mismatch.

PR #560, opened 2018-01-19
Locale Based Size Filter & DTS

billyjbryant:
## Description
This adds locale support for the <size> DTS, it also creates a "size_id" which can be used in filters as well as the size name. 

Size filters support all locales as viable filter options

## Type of Change

## Motivation and Context
Further extends multi-lingual support by allowing Size to be displayed per Locale

## How Has This Been Tested?
This was tested locally via Unit Testing and the Webhook Tester

2018-01-19 17_33_57- mod-chat - discord

## Wiki Update

## Checklist


PokeAlarm/Utilities/MonUtils.py
```diff
@@ -69,6 +69,29 @@ def get_gender_sym(gender):  # TODO - support other languages
...
+            return get_size_id.ids[name]
+        else:
+            return int(name)  # try as an integer
+    except Exception:
+        raise ValueError("Unable to interpret `{}` as a valid"
+                         " size name or id.".format(size_name))
+
+
 # Returns the (appraisal) size of a pokemon:
 def validate_pokemon_size(size_name):
```
kvangent: We can probably remove this now, right?

PR #563, opened 2018-01-21
Time Based Filters for Egg, Raid and Monster

billyjbryant:
## Description
This establishes time based filtering for Eggs, Raids and Monsters.  

## Type of Change

## Motivation and Context
Bringing time into the filter level eliminates the need for the `timelimit` configuration option.

## How Has This Been Tested?
This was tested with unit-tests

N/A

## Wiki Update

## Checklist

### DIFF
--- a/PokeAlarm/Events/MonEvent.py
+++ b/PokeAlarm/Events/MonEvent.py
@@ -7,7 +7,7 @@ from PokeAlarm.Utilities import MonUtils
 from PokeAlarm.Utils import (
     get_gmaps_link, get_move_damage, get_move_dps, get_move_duration,
     get_move_energy, get_pokemon_size, get_applemaps_link, get_time_as_str,
-    get_time_in_seconds, get_dist_as_str, get_weather_emoji)
+    get_dist_as_str, get_weather_emoji)
 from . import BaseEvent
 
 
@@ -25,7 +25,6 @@ class MonEvent(BaseEvent):
 
         # Time Left
         self.disappear_time = datetime.utcfromtimestamp(data['disappear_time'])
-        self.spawn_time_left = get_time_in_seconds(self.disappear_time)
 
         # Spawn Data
         self.spawn_start = check_for_none(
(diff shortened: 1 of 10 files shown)


PokeAlarm/Filters/EggFilter.py
```diff
@@ -21,6 +21,16 @@ def __init__(self, name, data):
             event_attribute='distance', eval_func=operator.ge,
             limit=BaseFilter.parse_as_type(float, 'max_dist', data))
 
+        # Time Left
+        self.min_time_to_hatch = self.evaluate_attribute(
+            # f.min_time_to_hatch <= r.hatch_time_left
+            event_attribute='hatch_time_left', eval_func=operator.le,
+            limit=BaseFilter.parse_as_type(int, 'min_time_to_hatch', data))
```
kvangent: The rest are all `min_time_left` and `max_time_left` - maybe this one should be as well?

PokeAlarm/Utils.py
```diff
@@ -437,6 +437,14 @@ def get_time_as_str(t, timezone=None):
     return time_left, time_12, time_24
 
 
+# Return the time in seconds
+def get_time_in_seconds(t, timezone=None):
```
kvangent: This should probably be called something like 'get_seconds_remaining' to make it more clear what it does.

It doesn't need timezone, so we should remove it.

PokeAlarm/Events/MonEvent.py
```diff
@@ -25,6 +25,7 @@ def __init__(self, data):
 
         # Time Left
         self.disappear_time = datetime.utcfromtimestamp(data['disappear_time'])
+        self.spawn_time_left = get_time_in_seconds(self.disappear_time)
```
kvangent: What do you think about changing all the event attributes to just 'time_left' so there is some consistency between them?

PokeAlarm/Utils.py
billyjbryant: Ok, i'll make that change.

PokeAlarm/Filters/EggFilter.py
billyjbryant: I thought about that when I was making it, was waiting for someone else to make the same suggestion. :)

PokeAlarm/Events/MonEvent.py
billyjbryant: 👌

PR #565, opened 2018-01-22
Adding Monster Type(s) DTS and Event Attribute

billyjbryant:
## Description
This adds the ability to include the Monster Type in alerts via DTS.  

The Types have been moved to Locales to provide multilingual support.  
The Base Stats and Boosted Weather data sets has been updated to reference the `type_id` instead of the `type_name`. 

This also sets up the necessary Event Attributes for Filtering; however, I could not get the filter to work in any way that made sense to me so this has not been done at this time. Will need to revisit that portion. 

### DTS 
The following DTS fields have been added to both **Monster** and **Raid** alerts

| DTS              | Description                                             |
|----------------- |-------------------------------------------------------- |
| type1            | The name of the Monster's Primary Type or `?`.          |
| type1_or_empty   | The name of the Monster's Primary Type or Empty.        |
| type2            | The name of the Monster's Secondary Type or `?`.        |
| type2_or_empty   | The name of the Monster's Secondary Type or Empty.      |
| types            | The Monster's Type formated as "type1/type2"            |

## Type of Change

## Motivation and Context
Allows to see the type of a Monster in alerts. 

## How Has This Been Tested?
Tested locally using the Webhook Tester

## Wiki Update

## Checklist


PokeAlarm/Utils.py
```diff
@@ -283,13 +283,19 @@ def get_base_types(pokemon_id):
...
-                }
+                get_base_types.info[int(id_)] = [
+                    j[id_].get('type1'),
+                    j[id_].get('type2')
+                ]
     return get_base_types.info.get(pokemon_id)
 
 
+# Returns the types for a pokemon
+def get_mon_type(pokemon_id):
```
kvangent: I think you can remove this

billyjbryant:
@AshKetchumBot can you advise which Emojis you would envision being used?

PR #570, opened 2018-01-24
Removing <boosted_weather_or_empty> DTS

billyjbryant:
## Description
`<boosted_weather>` can never be empty, so the `<boosted_weather_or_empty>` DTS is un-necessary.  Removing from code. 

## Type of Change

## Motivation and Context
@evilmoses requested someone to put in a change to fix this. 

## Wiki Update

## Checklist

### DIFF
--- a/PokeAlarm/Events/MonEvent.py
+++ b/PokeAlarm/Events/MonEvent.py
@@ -142,7 +142,6 @@ class MonEvent(BaseEvent):
             'weather_emoji': get_weather_emoji(self.weather_id),
             'boosted_weather_id': self.boosted_weather_id,
             'boosted_weather': boosted_weather_name,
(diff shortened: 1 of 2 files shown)


kvangent:
Can boosted weather be unknown? Do we care if it is?

evilmoses:
Why remove it? Is useful for some people...... The issue is just that the boosted_weather_name is never empty, as set to 'None', so the call to Unknown.or_empty(boosted_weather_name) will never return empty.

All that is needed is to test the boosted_weather_id and set to '' if 0, something like:

'boosted_weather_or_empty': "" if boosted_weather_id ==0 else boosted_weather_name etc, sorry unsure the correct syntax here :(

billyjbryant:
@evilmoses I must have mis-understood your changes you wanted to make.  I am going to close this PR and delete the branch.  Please PR what you want to change.

PR #571, opened 2018-01-24
boosted_or_empty DTS

LegitDongo:
## Description
This DTS -- `boosted_or_empty` will show the word `boosted` in the appropriate locale based on if the mon or raid mon is weather boosted.

This also has a basic Google Translated version of the "boosted" word locale

## Type of Change

## Motivation and Context
Not everyone knows the weather conditions to which a mon is boosted, thus having a clear and concise `boosted` word there would be best
Ex.
`A <boosted_or_empty> <mon_name> spawned!` would translate to `A boosted Dragonite spawned!`

## How Has This Been Tested?
Windows 10 using the webhook tester

## Wiki Update

## Checklist


locales/de.json
```diff
@@ -959,5 +959,6 @@
 	"5": "Windy",
 	"6": "Snow",
 	"7": "Fog"
-  }
+  },
+  "boosted": "verstärkt"
```
kvangent: Instead of leaving this out, what do you think of using a 'Misc' subsection for random words that we need translations for? I'm just worried this 'boosted' section will get lost if we process this file. Do you feel like this is a concern?

LegitDongo: Fine with me. I was kinda thinking that when I made it anyway 😄

evilmoses:
Why call is_weather_boosted again? This has already been done so no need to call again....

'boosted_or_empty': locale.get_boosted_text() if
             is_weather_boosted(self.mon_id, self.weather_id) else ''

Just need to check boosted_weather_id and call locale.get_boosted_text() if 0. This applies to raid event and mon events.

evilmoses:
I just submitted a fix to another issue similar to this. I used this code:

'' if self.boosted_weather_id == 0 else boosted_weather_name.

You could use:

'boosted_or_empty': locale.get_boosted_text() if
             self.boosted_weather_id  == 0 else '',

etc etc

kvangent:
This will need to be updated for changing in #572 as well.

PR #572, opened 2018-01-24
Fix boosted_weather_or_empty

evilmoses:
## Description
Currently boosted_weather_or_empty never returns empty, as the boosted_weather_name is never empty. This PR fixes that

## Type of Change

### DIFF
--- a/PokeAlarm/Events/MonEvent.py
+++ b/PokeAlarm/Events/MonEvent.py
@@ -144,7 +144,7 @@ class MonEvent(BaseEvent):
             'weather_emoji': get_weather_emoji(self.weather_id),
             'boosted_weather_id': self.boosted_weather_id,
             'boosted_weather': boosted_weather_name,
-            'boosted_weather_or_empty': Unknown.or_empty(boosted_weather_name),
+            'boosted_weather_or_empty': '' if self.boosted_weather_id == 0 else boosted_weather_name,
             'boosted_weather_emoji':
                 get_weather_emoji(self.boosted_weather_id),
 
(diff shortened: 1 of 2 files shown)


evilmoses:
That is to replicated the functionality of the Mon boosted weather where value is always sent, but as 0 when not boosted, hence doing it in the raids. I would really rather prefer the functionality were kept the same even if we are just replicating. Happy to have another id_or_empty (though god knows who would use it?!)

evilmoses:
Oh wait, you mean if scanner isn't currently sending it?

kvangent:
Yes - if we are going to add a 'or_empty' variant, the original should show 'unknown' when PA has no idea since it shows 'None' when it's not boosted.

evilmoses:
The original? Original what sorry? Not boosted <> No data. As long as it is clear the difference between these two, no issues. I see what you say, that right now, there is no difference for raids as boosted_weather_id is set to 0 initially, but need to ensure that the distinction is clear. Will give some thought and update. If no data sent, set to '?' ??

kvangent:
`boosted_weather_id` should default to '?' if not provided in the webhook. `boosted_weather` should then be 'unknown' and `boosted_weather_or_empty` should be an empty string if `boosted_weather` is unknown or None.

kvangent:
Fixed in https://github.com/PokeAlarm/PokeAlarm/pull/571/commits/d2beaf6144f9abf1840ae33d67c4867197487257

kvangent:
Closing - feel free to comment if it's still not working correctly.

PR #579, opened 2018-01-27
Conversion tool for 3.5 to 3.6 alarms.

kvangent:
This adds a tool which allows the conversion from the v3.5 format into the v3.6 format. Telegram users will want to verify their settings, as it only replaces keys and doesn't change the settings themselves.

### DIFF
--- a/tools/convert_alarms_file.py
+++ b/tools/convert_alarms_file.py
@@ -72,5 +72,4 @@ if __name__ == '__main__':
         exit(1)
 
     print ("Conversion finished! If you are a telegram user, verify your"
-          "settings as some of them have inversed.")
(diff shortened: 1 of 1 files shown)


tools/convert_alarms_file.py
```diff
@@ -0,0 +1,75 @@
...
+from collections import OrderedDict
+
+
+def exchange(key):
+    return {
+        "pokemon": "monsters",
+        "pokestop": "stops",
+        "gym": "gyms",
+        "egg": "eggs",
+        "gym": "gyms",
```
billyjbryant: You missed `raid` -> `raids`

PR #594, opened 2018-02-02
Weather Fixes

evilmoses:
## Description
This PR fixes some issues in the weather alerts for mons and raids:

1. MonEvent - Changed back to setting boosted weather to 0. This is because if mon is not weather boosted scan sends None to webhook, so there is no way to differentiate between sent and not boosted and not sent. Having it set to ? mean all unboosted mons were posting unknown instead of 'None'.
1. There is an unnecessary comma that appends to a string, meaning later name lookups fails.

```
boosted_weather_id = \
            0 if Unknown.is_not(self.weather_id) else Unknown.TINY,
```

The comma at the end was causing issues. I have also tidied up the code to make it similar to MonEvent

## Type of Change

### DIFF
--- a/PokeAlarm/Events/MonEvent.py
+++ b/PokeAlarm/Events/MonEvent.py
@@ -44,7 +44,7 @@ class MonEvent(BaseEvent):
         self.weather_id = check_for_none(
             int, data.get('weather'), Unknown.TINY)
         self.boosted_weather_id = check_for_none(
-            int, data.get('boosted_weather'), Unknown.TINY)
+            int, data.get('boosted_weather'), 0)
 
         # Encounter Stats
         self.mon_lvl = check_for_none(
--- a/PokeAlarm/Events/RaidEvent.py
+++ b/PokeAlarm/Events/RaidEvent.py
@@ -32,14 +32,22 @@ class RaidEvent(BaseEvent):
         self.lng = float(data['longitude'])
         self.distance = Unknown.SMALL  # Completed by Manager
         self.direction = Unknown.TINY  # Completed by Manager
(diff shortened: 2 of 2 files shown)


kvangent:
Which scanner is sending a null value instead of '0'? Why is there the inconsistency between `boosted_weather` and `weather`?

locales/zh_hk.json
```diff
@@ -6,31 +6,31 @@
...
+    "13": "\u96fb",
+    "14": "\u8d85\u80fd\u529b",
     "15": "\u51b0",
-    "16": "\u9f99",
-    "17": "\u4ed9\u5973",
-    "18": "\u9ed1\u6697"
+    "16": "\u9f8d",
+    "17": "\u60e1",
+    "18": "\u5996\u7cbe"
   },
```
billyjbryant: Looks like you have a bad rebase, your PR includes commits for other merges.

evilmoses:
Boosted weather value for pokemon, it appears as though it sends the boosted weather or nothing if not boosted.

Weather of course, always has a value.

locales/zh_hk.json
evilmoses: I did nothing more than pull from Dev into new branch and make changes. I didn't rebase anything. Why would pulling latest not include change?

billyjbryant: When did you pull from Dev to create your branch? Your PR includes 61c81e2 as a commit which was merged into Dev 3 days ago.

kvangent: There was a commit in dev that wasn't in master. I've corrected the issue and rebased the branch since it was a mistake on my part.

PR #598, opened 2018-02-07
Bug Fix: Timezone Incorrect on Stops

LegitDongo:
## Description
Fixes stops not being in the correct timezone

## Type of Change

## Motivation and Context
Things relating to time weren't in the correct timezone

## How Has This Been Tested?
with guy with the bug (and it's kind of obvious that it's broke based off code)

## Wiki Update

## Checklist


LegitDongo:
Didn't realize that I hadn't changed my commit message from the last thing I did 😅

PR #599, opened 2018-02-07
Make it possible to add a list of monsters to NOT trigger alarm.

fosJoddie:
## Description
Implements a monster blacklist that will make filters fail for certain pokemon.

Example use: 

Alert about any pokemon above 15/14/14 iv, unless they are certain "trash" pokemon:

```
"f_15_14_14_ignore_trash":{
"ignore_monsters": [ 10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,27 ... ],
   "min_atk": 15, "max_atk": 15,
   "min_def": 14, "max_def": 15,
   "min_sta": 14, "max_sta": 15,
   "min_iv": 0.0, "max_iv": 100, 
   "is_missing_info": false
}
```

## Type of Change

## Motivation and Context
Instead of writing a huge list of pokemon I WANT alerts on, I wanna write a smaller list of pokemon I do not want.

## How Has This Been Tested?
I wrote a python parser on the bathroom stall at work and ran my code through it there. It seems to work fine until it runs out of toilet paper.

## Wiki Update
Do you still use the wiki? I wrote up the change in the monster filter docs file

## Checklist


billyjbryant:
What happens if you have both a `"monsters"` list and an `"ignore_monsters"` list in an incident that they conflict with each other?

billyjbryant:
You should also look at creating the necessary tests within the unit testing, this will prove that the filters work as expected and satisfies the "How Has This Been Tested?" requirement of the PR

fosJoddie:
If something is in the ignore filter, it will fail the filter for that pokemon. 
Blacklist takes precedence over Whitelist in this case, simply because of the way the filter logic works

tests/filters/test_monster_filter.py
```diff
@@ -41,6 +41,31 @@ def test_monster_id(self):
         for e in [fail1, fail2, fail3]:
             self.assertFalse(mon_filter.check_event(e))
 
+    def test_ignore_monster_id(self):
+        # Create the filters
+        settings = {
+            "monsters": [1, "2", "Venusaur"],
```
billyjbryant: For testing here, I would remove `"monsters"` as we truly just want to test the ignore filter in this particular scenario.

fosJoddie: sure thing, updated

kvangent:
Do we need to add a similar filter to raids?

wickles:
What is the use case for this? Temporary ignore lists for events maybe?

kvangent:
@wickles it's a blacklist for monsters

darkelement1987:
I liked it the  way it was on older version of pokealarm (true/false)

PR #603, opened 2018-02-12
Add restrictions and DTS for Gym Sponsor and Parks.

evilmoses:
## Description
If scanner sends the gym sponsor and park details (park = name of park in OSM qualifying gym for EX raid) this PR adds a DTS for them, and ability to filter on these fields.

## Type of Change

## Checklist


kvangent:
What exactly are 'sponser' and 'park' fields? I take it this is Monocle only? 

I see you've added fields to the webhook tester, but it doesn't look like any questions to edit those fields.

You'll also need to add unit tests for the new restrictions.

LegitDongo:
Here's some info on RMs progress with EX gym detection
https://github.com/RocketMap/RocketMap/pull/2440
From the looks of the PR, they don't plan on adding it to webhooks like in this PR

bbdoc:
Monocle (Hydro's fork) does indeed have a mechanism to detect if a gym is in the same s2 cell as a park and store it in DB (forts table). It also insert in DB the information that a gym is a sponsored one. This allows us to show the information on our maps (typically I do it using PMSF). Sending it through webhook allows to easily make the difference wether a gym is EX-eligible or not and send this information in our alarms. 

In my case, I have a filter using `"gym_park_contains": [ ".*" ]` in association with `"is_missing_info": false` which creates a custom dts: `"custom_dts": { "EX": "\n\u26A0 *Potential EX Gyms* \u26A0" }` which I then can use in my alarms.

kvangent:
So the `"park"` field is the name of the OSM park - what happens if there is no OSM park? Is it any empty string or null?

What is 'sponsored' then? It seems to be an int, but the filter is by regex? Why not just use a set?

KartulUdus:
@LegitDongo I do plan to add it to webhook; however, I'd like to get 2440 in first.
Don't know about sponsored gyms, we don't have any ☹️

evilmoses:
"You'll also need to add unit tests for the new restrictions." - Que?!

kvangent:
We have unit tests for the filters to ensure they are working correctly located in the `tests/` folder. You can use `python -m unittest discover` from the root folder to run them all.

evilmoses:
[xxxxx@xxxxx PokeAlarm]$ python -m unittest discover
..................................................
----------------------------------------------------------------------
Ran 50 tests in 0.107s

OK

PokeAlarm/Events/EggEvent.py
```diff
@@ -44,6 +44,10 @@ def __init__(self, data):
             str, data.get('description'), Unknown.REGULAR).strip()
         self.gym_image = check_for_none(
             str, data.get('url'), Unknown.REGULAR)
+        self.gym_sponsor = check_for_none(
+            int, data.get('sponsor'), Unknown.SMALL)
```
kvangent: what is the `sponsor` field, and how is it determined?

PokeAlarm/Filters/EggFilter.py
```diff
@@ -46,6 +47,18 @@ def __init__(self, name, data):
             limit=BaseFilter.parse_as_set(
                 GymUtils.create_regex, 'gym_name_contains', data))
 
+        # Gym sponsor
+        self.gym_sponsor_index_contains = self.evaluate_attribute(
+            event_attribute='gym_sponsor', eval_func=GymUtils.match_regex_dict,
```
kvangent: `'gym_sponsor'` is parsed as an int - why are we comparing as a regex instead of a set of ints (like monster_id?)

PokeAlarm/Filters/EggFilter.py
```diff
@@ -46,6 +47,18 @@ def __init__(self, name, data):
             limit=BaseFilter.parse_as_set(
                 GymUtils.create_regex, 'gym_name_contains', data))
 
+        # Gym sponsor
```
kvangent: You need to add new unit tests when adding a new restriction to a filter.

tools/webhook_test.py
```diff
@@ -126,6 +126,8 @@ def set_init(webhook_type):
             "message": {
                 "gym_id": 0,
                 "gym_name": "unknown",
+                "park": None,
+                "sponsor": 4,
```
kvangent: should probably default this to none

tools/webhook_test.py
```diff
@@ -137,6 +139,8 @@ def set_init(webhook_type):
             "message": {
                 "gym_id": 0,
                 "gym_name": "unknown",
+                "park": "Test Park",
```
kvangent: These should default to None as well.

PokeAlarm/Events/EggEvent.py
evilmoses: Sponsor is the sponsor flag for a gym, as in the companies sponsoring gyms. We do not have any in UK so cannot verify how this works - this was originally a PR to my fork, that I guessed would be useful on stock.

PokeAlarm/Filters/EggFilter.py
evilmoses: I do not know sorry. Code was taken verbatim from my fork where users have been using it for weeks, tidied up to appease Travis and submitted. Is it better to leave out the sponsor stuff and just submit the EX park stuff instead? As far as I know, nobody is using sponsor, but lots are using the park stuff to filter on EX raid eligibility

tools/webhook_test.py
evilmoses: I might remove based on above comments....

PokeAlarm/Filters/EggFilter.py
evilmoses: The EX raid stuff is badly needed, the sponsor stuff less so.

PokeAlarm/Filters/EggFilter.py
evilmoses: Can you elaborate?

PokeAlarm/Events/EggEvent.py
billyjbryant: what are the possible values of the sponsor field?  Is this in Proto?

PokeAlarm/Filters/EggFilter.py
billyjbryant: I believe Sponsor is valuable (at least in the US) as the majority of EX Raids are @ Sponsored Gyms in the US.   So I'd say both are equally important.  

I do think that if `sponsor` is just an id number, it would be better to compare this as int instead of as a regex.  I can help with converting this if you'd like

PokeAlarm/Filters/EggFilter.py
billyjbryant: Look in `tests/filters/test_egg_filter.py` and `tests/filters/test_raid_filter.py`, these files create the unit testing for all Raid & Egg Filters.  This is what @kvangent is wanting you to create, add in a unit test to test that your filters work.  

Once you have created the unit test, you can test it by running `python -m unittest discover`

PokeAlarm/Events/EggEvent.py
evilmoses: Yes, looks like 23 is current max value used

evilmoses:
Having reviewed this code myself (I didn't use it and therefore didn't need to understand it) and seen that the sponsor is indeed an integer, why are we regexing it? My guess is that all we need to filter on is whether it is a sponsored gym or not. I cannot see the need to see which sponsor it is etc.

Therefore I've reworked this to use a simple boolean for the filter for sponsors - is_sponsor. 

It has passed the unit tests and travis now, though maybe by pure luck.....

PokeAlarm/Events/EggEvent.py
kvangent: @evilmoses Is it a value derived from the pgo api or is it determined by the scanner?

evilmoses: Pogo protos

kvangent: If we are going to this 'bool' - Python considers '0' to be 'falsey', so that should give us False for 0 and True for everything else. ('None' will still be treated as unknown).

In the future someone may want to expand upon this to filter by specific sponsor, but I'm fine with this for now.

PokeAlarm/Events/RaidEvent.py
```diff
@@ -176,6 +180,8 @@ def generate_dts(self, locale, timezone, units):
             'gym_name': self.gym_name,
             'gym_description': self.gym_description,
             'gym_image': self.gym_image,
+            'gym_sponsor': self.gym_sponsor,
```
kvangent: If we are going to leave this in, we should probably change to something like `is_sponsored` so it's clear it will be a True/False value.

PokeAlarm/Filters/RaidFilter.py
```diff
@@ -126,6 +138,14 @@ def to_dict(self):
         if self.gym_name_contains is not None:
             settings['gym_name_matches'] = self.gym_name_contains
 
+        # Gym Sponsor
+        if self.is_sponsor is not None:
+            settings['gym_is_sponsor'] = self.is_sponsor
```
kvangent: Update to `is_sponsor` to match the attribute name.

tests/filters/test_egg_filter.py
```diff
@@ -174,7 +215,9 @@ def generate_egg(values):
         "end": 1499246052,
         "level": 5,
         "latitude": 37.7876146,
-        "longitude": -122.390624
+        "longitude": -122.390624,
+        "sponsor": 4,
```
kvangent: Default to `None`

tests/filters/test_egg_filter.py
```diff
@@ -174,7 +215,9 @@ def generate_egg(values):
...
+        "gym_park": "Test Park"
```
kvangent: Default to `None`

tools/webhook_test.py
```diff
@@ -125,7 +125,10 @@ def set_init(webhook_type):
             "type": "raid",
             "message": {
                 "gym_id": 0,
-                "gym_name": "unknown",
+                "name": "Test gym",
+                "team": 1,
+                "park": None,
+                "sponsor": 0,
```
kvangent: This should probably default to `None`

tools/webhook_test.py
```diff
@@ -136,15 +139,18 @@ def set_init(webhook_type):
             "type": "raid",
             "message": {
                 "gym_id": 0,
-                "gym_name": "unknown",
+                "name": "Test gym",
+                "team": 1,
```
kvangent: Some of these changes are going to cause some unintended consequences. It might be best to just remove these changes all together - but I'll leave that up to @LegitDongo since the webhook tester is his baby.

kvangent:
Also this is in conflict at the moment, so it'll need to be rebased before Travis will run tests on it.

PokeAlarm/Events/RaidEvent.py
evilmoses: That won't be a true or false value. That will be the integer representing the sponsor. It is only the filter that would be true or false

PokeAlarm/Events/EggEvent.py
evilmoses: Um, that's as intended. 0 is false, >0 is true and None is unknown. For non sponsored gyms the value is always 0. In the events though, we will always return into, not bool. Book is for the filter only.

PokeAlarm/Filters/RaidFilter.py
evilmoses: I made it different on purpose as all the other values here are different, despite having no idea why?!

evilmoses:
The DTS will supply the sponsor integer value, not that this is useful info really. The filter is a Boolean to determine whether the gym is a sponsored gym or not. Sponsor is always in the fort info, but set to 0 when not sponsored, so available values are 0 for not sponsored, > 0 if sponsored, or None if not sent by scanner.

kvangent:
That's my point - the ID is kinda worthless. We should either treat it as a boolean and expand it in the future, or else we need to map the ids to the sponsor name so that it returns the name of the sponsor.

evilmoses:
Ah ok, get you.

evilmoses:
I'll change the event to also output a Boolean too.

evilmoses:
We don't have the names of the sponsors to my knowledge, and it would need updates every time a new one was added for little benefit.

tools/webhook_test.py
```diff
@@ -126,6 +126,9 @@ def set_init(webhook_type):
             "message": {
                 "gym_id": 0,
                 "gym_name": "unknown",
+                "team": 0,
```
kvangent: We need to leave `team` out until we get better support added for it - otherwise it could effect the value in gym cache. 

I think we should just drop all changes to the webhook tester for now, so that they will just show up as missing info (unless you have a better suggestion)

PokeAlarm/Events/EggEvent.py
```diff
@@ -44,6 +44,10 @@ def __init__(self, data):
             str, data.get('description'), Unknown.REGULAR).strip()
         self.gym_image = check_for_none(
             str, data.get('url'), Unknown.REGULAR)
+        self.gym_sponsor_id = check_for_none(
```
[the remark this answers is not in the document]
evilmoses: Disagree, keeps same naming structure as other gym fields.

PokeAlarm/Events/EggEvent.py
```diff
@@ -44,6 +44,10 @@ def __init__(self, data):
...
+            int, data.get('sponsor'), Unknown.TINY)
+        self.gym_park = check_for_none(
```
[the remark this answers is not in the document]
evilmoses: Disagree, keeps same naming structure as other gym fields.

evilmoses:
I've reworked this again.

gym_is_sponsor is now an event field, so that it can be used as a DTS. This is because the sponsor_id is basically useless as an output as discussed above, and using a boolean may be of more use.

This therefore means that the filters could be reworked to just use this output value instead, so functions have been removed and filters tidied.

flake8 and unittests passed.

PR #606, opened 2018-02-13
Caching Improvements

kvangent:
This is a bunch of miscellaneous cache improvements.

* Updated API to be more condensed
* Separated gym_info into gym_name, gym_desc, and gym_url into separate endpoints
* Added support for `team_id` from RM egg and raid webhooks
* FileCache now creates the cache folder upon initialization if it doesn't exist
* FileCache saves initial file with the ".new" extension and then renames over the original file to reduce the chance of the cache being corrupted if PA is closed incorrectly
* If FileCache fails to load the file for any reason, it will continue on as normal and overwrite it the cache on the next opportunity

### DIFF
--- a/PokeAlarm/Cache/Cache.py
+++ b/PokeAlarm/Cache/Cache.py
@@ -3,7 +3,8 @@ import logging
 from datetime import datetime
 # 3rd Party Imports
 # Local Imports
-from ..Utils import get_image_url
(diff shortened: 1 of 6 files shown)


billyjbryant:
Pulled this, testing now.

billyjbryant:
Pulled, testing.

PR #608, opened 2018-02-15
bugfix: If the only address DTS you use is `<address_eu>` reverse geocode fails

fosJoddie:
## Description
if your alarm DTS only contains address_eu , geocode lookup won't happen and the alarm will become <address_eu>

## Type of Change

## Motivation and Context
It makes alarms great again also in parts of the world that likes to write `street streetnumber`

## Wiki Update

## Checklist


kvangent:
The auto-detect and enable API's is probably going away soon, but if you really want I can pull this.

fosJoddie:
If its going bye bye I don’t mind, I just noticed it was missing after removing city from my own configs. Up to you.

PR #610, opened 2018-02-18
Move Types

billyjbryant:
## Description
Adds support for Move Types to the move data files.  Also adds in Move Type DTS fields. 

### DTS 
The following DTS fields have been added to both **Monster** and **Raid** alerts

| DTS              | Description                                             |
|----------------- |-------------------------------------------------------- |
| quick_type_id | The id of the Quick Move's Type or  `?`.          |
| quick_type     | The name of the Quick Move's Type or `Unknown`.        |
| quick_type_emoji            | The emoji for the Quick Move's Type or Empty.        |
| charge_type_id | The id of the Charge Move's Type or  `?`.          |
| charge_type     | The name of the Charge Move's Type or `Unknown`.        |
| charge_type_emoji            | The emoji for the Charge Move's Type or Empty.        |

## Type of Change

## Motivation and Context
Knowing the type of a move for a monster or raid boss is beneficial when planning raid groups and building effective battle strategies.  

## How Has This Been Tested?
This was tested on my production instance

image

image

## Wiki Update

## Checklist

### DIFF
--- a/PokeAlarm/Events/MonEvent.py
+++ b/PokeAlarm/Events/MonEvent.py
@@ -202,9 +202,9 @@ class MonEvent(BaseEvent):
             'quick_move': locale.get_move_name(self.quick_id),
             'quick_id': self.quick_id,
             'quick_type_id': self.quick_type,
-            'quick_type': locale.get_type_name(self.quick_type),
+            'quick_type': locale.get_type_name(self.quick_type_id),
             'quick_type_emoji': Unknown.or_empty(
-                get_type_emoji(self.quick_type)),
+                get_type_emoji(self.quick_type_id)),
             'quick_damage': self.quick_damage,
             'quick_dps': self.quick_dps,
             'quick_duration': self.quick_duration,
@@ -214,9 +214,9 @@ class MonEvent(BaseEvent):
             'charge_move': locale.get_move_name(self.charge_id),
             'charge_id': self.charge_id,
             'charge_type_id': self.charge_type,
-            'charge_type': locale.get_type_name(self.charge_type),
+            'charge_type': locale.get_type_name(self.charge_type_id),
             'charge_type_emoji': Unknown.or_empty(
-                get_type_emoji(self.quick_type)),
+                get_type_emoji(self.charge_type_id)),
             'charge_damage': self.charge_damage,
             'charge_dps': self.charge_dps,
             'charge_duration': self.charge_duration,
--- a/PokeAlarm/Events/MonEvent.py
+++ b/PokeAlarm/Events/MonEvent.py
@@ -5,7 +5,7 @@ from datetime import datetime
 from PokeAlarm import Unknown
 from PokeAlarm.Utilities import MonUtils
 from PokeAlarm.Utils import (
-    get_gmaps_link, get_move_damage, get_move_dps,
+    get_gmaps_link, get_move_type, get_move_damage, get_move_dps,
     get_move_duration, get_move_energy, get_pokemon_size,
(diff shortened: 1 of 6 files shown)


billyjbryant:
I found an issue with this, it looks to be using the same ID for both.

billyjbryant:
Fixed the issue with the Emoji and provided better screenshots.

data/move_info.json
```diff
@@ -1,1115 +1,1301 @@
 {
   "013": {
+	"type": 1,
```
MARTINI1: tabulate issue

data/move_info.json
```diff
@@ -1,1115 +1,1301 @@
...
     "damage": 60,
     "dps": 20.69,
     "duration": 2900,
     "energy": 33
-  },
+  },
```
MARTINI1: another tabulate that is left over

PokeAlarm/Events/MonEvent.py
```diff
@@ -196,16 +199,24 @@ def generate_dts(self, locale, timezone, units):
             'form_id_3': "{:03d}".format(self.form_id),
 
             # Quick Move
-            'quick_move': locale.get_move_name(self.quick_move_id),
-            'quick_id': self.quick_move_id,
+            'quick_move': locale.get_move_name(self.quick_id),
+            'quick_id': self.quick_id,
+            'quick_type_id': self.quick_type,
+            'quick_type': locale.get_type_name(self.quick_type),
+            'quick_type_emoji': Unknown.or_empty(
```
kvangent: Doesn't type emoji return `''`? If so, the unknown check is superfluous

PokeAlarm/Events/MonEvent.py
```diff
@@ -196,16 +199,24 @@ def generate_dts(self, locale, timezone, units):
...
             'quick_energy': self.quick_energy,
 
             # Charge Move
-            'charge_move': locale.get_move_name(self.charge_move_id),
-            'charge_id': self.charge_move_id,
+            'charge_move': locale.get_move_name(self.charge_id),
+            'charge_id': self.charge_id,
+            'charge_type_id': self.charge_type,
+            'charge_type': locale.get_type_name(self.charge_type),
+            'charge_type_emoji': Unknown.or_empty(
```
kvangent: Same about Unknown check

PokeAlarm/Utils.py
```diff
@@ -128,6 +128,18 @@ def get_team_id(team_name):
...
+# Returns the types of a move when requesting
+def get_move_type(move_id):
+    if not hasattr(get_move_type, 'info'):
+        get_move_type.info = {}
+        file_ = get_path('data/move_info.json')
+        with open(file_, 'r') as f:
+            j = json.loads(f.read())
+        for id_ in j:
+            get_move_type.info[int(id_)] = j[id_]['type']
+    return get_move_type.info.get(move_id, 'unkn')
```
kvangent: We should hardcode this Unknown using the unknown class

billyjbryant: Do you want me to do the same for the rest of the Move Data while I'm in here or submit that via another PR?

kvangent: That can be in another PR - I think that we need to change our locales change to a singleton now that we removed multiprocessing. There's no reason to keep an identical copy of the mappings per manager.

billyjbryant: I also think a lot of the functions in `Utils.py` really need to be moved to `MonUtils.py` so that can be another cleanup PR

PR #612, opened 2018-02-20
GMaps Improvements

kvangent:
## Description

This is a rewrite of the Google Maps API calls used for DTS. It removes the official google python client and instead handles the requests directly to give more control over how they are handled. This will allow for better error handling and won't timeout for extended periods of time when the API Quota has been reached.

Additionally, it removes the automatic DTS detection and instead adds explicit config options to enable the use of DTS. This is to reduce issues users are experiencing with accidentally overusing their quota. (as a side effect I've cleaned up the config.ini.example as well)

Some of the Distance Matrix DTS have been renamed, and are as follows:
* `<walking_distance>` and `<walking_duration>`
* `<biking_distance>` and `<biking_duration>`
* `<driving_distance>` and `<driving_duration>`
* `<transit_distance>` and `<transit_duration>` - (these two are new)

## Type of Change

- [X] Breaking change (would cause existing functionality to change)

## Motivation and Context

The official GMaps client repeatedly spams the API server until it times out, which can cause several seconds of blocking when trying to trigger a notification. Additionally, some users were hitting their quotas by accident because they were using APIs they didn't know they had enabled.

## Wiki Update


The-Real-Ketchum-Dev:
The new config file is amazing gonna try this out right now

The-Real-Ketchum-Dev:
Works like a charm! Great config.ini and my key isn't spiking anymore

elincognito:
I keep getting
`     parse_settings(os.path.abspath(os.path.dirname(__file__)))
   File "start_pokealarm.py", line 287, in parse_settings
     debug=config['DEBUG']
   File "/root/rm2/PokeAlarm/PokeAlarm/Manager.py", line 61, in __init__
     self.set_location(location)
   File "/root/rm2/PokeAlarm/PokeAlarm/Manager.py", line 489, in set_location
     self.__location = self._gmaps_service.get_location_from_name(
 AttributeError: 'GMaps' object has no attribute 'get_location_from_name'`

config/config.ini.example
```diff
@@ -1,27 +1,76 @@
...
+
+
+#########################
+# Manager Settings
+#########################
+# All of the settings below this line are manager-specific.
+# If a single setting is supplied, it will apply to all Managers.
+# Example: `locale: en` will set all Managers to english
+# If an array of settings is supplied, they will be apply to Managers in order.
+# Example: `locale: [ en, fr, sp ]` sets a different langauge for 3 Managers.
```
MARTINI1: `sp` does not exist, I think you want to say spanish so the correct option would be `es`

config/config.ini.example
```diff
@@ -1,27 +1,76 @@
...
+
+
+########################
+# Webserver Settings
+########################
+
+#host: 127.0.0.1                # Interface to listen on (default='127.0.0.1')
+#port: 4000						# Port to listen on (default='4000')
+#concurrency: 200               # Maximum concurrent connections to webserver (default=200)
+#manager_count: 1				# Number of Managers to run (default=1)
```
MARTINI1: I think this should go in the manager settings section

config/config.ini.example
```diff
@@ -1,27 +1,76 @@
...
+
+#filters: filters.json          # Filters for the Manager (default='filters.json')
+#alarms: alarms.json            # Alarms for the Manager (default='alarms.json')
+#rules: rules.json              # Rules for the Manager (default=None)
+#geofence: geofence.txt         # Geofences to be used in Filters (default=None)
+
+
+# Location Specific
+#####################
+location:                       # Location, as address or coordinates (default=None)
```
MARTINI1: this option not commented with #

config/config.ini.example
```diff
@@ -1,27 +1,76 @@
...
-#location:										# Location for the manager. 'Name' or 'lat lng' (default: None)
-#locale:										# Language to be used to translate names (default: en)
-#cache_type:                                    # Specify the type of cache to use. Options: ['mem', 'file'] (Default: 'mem')
-#unit:											# Units used to measure distance. Either 'imperial' or 'metric' (default: imperial)
-#timelimit:										# Minimum number of seconds remaining to send a notification (default: 0)
-#max_attempts:									# Maximum number of attempts an alarm makes to send a notification. (default: 3)
-#timezone:                                      # Timezone used for notifications Ex: 'America/Los_Angeles' or '[America/Los_Angeles, America/New_York]'
\ No newline at end of file
+# By default, PA will use `config/config.ini` to load settings.
+# You can override the config file selction with `--config-file` or `-cf`.
```
MARTINI1: `selction` -> `selection`

MARTINI1:
Besides the comments I think you have forgotten to close the section titles as webserver or manager

like this:
```
########################
# Webserver Settings
########################
```

The-Real-Ketchum-Dev:
It has broke the custom_dts for me

config/config.ini.example
elincognito: that explains my error above didn't notice 👀

kvangent:
Added regular geocoding and surrounded functions with locks to prevent redundant calls from rules, and addressed @MARTINI1's feedback.

@elincognito The error has been cleaned up, so it should be easier to tell what went wrong.

@The-Real-Ketchum-Dev I wasn't able to replicated the issue with custom_dts. Are you sure you have it set up correctly?

The-Real-Ketchum-Dev:
@kvangent unless something has changed with custom_dts then no.. It worked before adding this PR

config/config.ini.example
kvangent: I left it there because technically it is a web-server setting, instructing it how many Managers to spin up. It also doesn't take an array, and if I moved it any lower it would be under the line that says "Anything beneath this line takes multiple arguments."

kvangent:
@The-Real-Ketchum-Dev Please confirm the same config works on the current `dev` branch, and then switch to this branch and verify it breaks when you have a chance.

The-Real-Ketchum-Dev:
Will do.. I'm currently at work but will do so when i get home

MARTINI1:
changes added in wiki

PR #613, opened 2018-02-20
Adds Costume Support to MonEvent & RaidEvent

billyjbryant:
## Description
This PR adds support for the `costume` webhook field to both Monster & Raid Events.  It also adds locale and DTS similar to `form` as well as a `costume_ids` filter.  

The PR also starts the work of building the ability to filter by Form & Costume name in addition to ID. 

## Type of Change

## Motivation and Context
Adds support for the costume webhook recently added to RM

## How Has This Been Tested?
Through unit testing only at this moment

N/A

## Wiki Update

## Checklist


PokeAlarm/Filters/MonFilter.py
```diff
@@ -93,6 +93,10 @@ def __init__(self, name, data):
         self.forms = self.evaluate_attribute(  # f.forms in m.form_id
             event_attribute='form_id', eval_func=operator.contains,
             limit=BaseFilter.parse_as_set(int, 'form_ids', data))
+        # Costume  TODO: names
```
kvangent: Looks like you finished by name below?

tests/filters/test_monster_filter.py
```diff
@@ -201,6 +201,21 @@ def test_form(self):
         self.assertFalse(mon_filter.check_event(create_event({'form': 2})))
         self.assertFalse(mon_filter.check_event(create_event({'form': 999})))
 
+    def test_costume(self):
```
kvangent: Both this and the forms should be updated to use name

billyjbryant: Tried this and failed, leaving as ID for now.

PokeAlarm/Filters/MonFilter.py
billyjbryant: Nope

kvangent:
It looks like this needs a rebase.

Issue #595, opened 2018-02-03
[ENHANCEMENT] Add toggle to enable web page preview in Telegram messages


Alexus230:
## Description
From patch 3.6 PokeAlarm sends all messages to Telegram with option 'disable_web_page_preview'. Please add toggle to enable web page preview in Telegram messages.

## Why is this enhancement necessary or useful?
In my setup I use web page preview feature to include pokemon image inside a message. This trick help me reduce the amount of messages, because when stickers was enabled, I got 3 messages (sticker,message,map) on every event, but with this trick I got only 2 messages (message,map).

kvangent:
@Alexus230 #620

PR #620, opened 2018-02-24
Telegram Web Preview Option

LegitDongo:
## Description
#595 requested that the `disable_web_page_preview` option in the Telegram API be brought back
It makes sense for it to be configurable

## Type of Change

## Motivation and Context
Options

## How Has This Been Tested?
Locally on Windows 10

## Wiki Update

## Checklist


MARTINI1:
web_preview added in telegram wiki

PR #624, opened 2018-02-28
Dev Add new DTSs for Height and Weight

Romfrosk:
## Description
Add addtional weight & height DTSs matching IV style, with 0,1 & 2 decimals

## Type of Change

## Wiki Update

## Checklist


PokeAlarm/Events/MonEvent.py
```diff
@@ -252,6 +252,7 @@ def generate_dts(self, locale, timezone, units):
             'gender': self.gender,
             'height': self.height,
             'weight': self.weight,
+            'weight_short': "{:.1f}".format(self.weight),
```
kvangent: This will throw an error if weight is unknown

Romfrosk:
Changed according to comments from Kvangent
Thanks for pointing out the "break"

PR #626, opened 2018-03-01
Dev: Move Rejected messages to Debug

Romfrosk:
## Type of Change

## Wiki Update

## Checklist


kvangent:
The choice to be verbose with logging is intentional - the most often asked question is almost always some variation of "why didn't the monster I saw on the map get notified!?". 

However with the new system this can be fairly overwhelming since it shows a reject for _every_ filter that is checked against. We have a bunch of changes planned in the next patch (3.8) all centered around improving the logging experience for PA. One of these is going to be the ability to specify how verbose you want the logs to be.

I'm going to close this because we want to leave the verbosity on by default, but please know that we understand it's an issue and will be addressing it shortly.

PR #649, opened 2018-03-16
Support for Dynamic Rarity

kvangent:
This is support for scanners that send Dynamic Rarity through webhooks.

Expected data should be sent via the field `rarity` as an integer.

Currently there are 5 rarities:
`0` => `Common`
`1` => `Uncommon`
`2` => `Rare`
`3` => `Very Rare`
`4` => `Ultra Rare`

This adds two dts - `rarity_id` and `rarity`.

It also adds the restriction `rarity`, which expects an array of either names/ids. Example:
```json
"rarity": [ "common", "uncommon", "rare", "very rare", "ultra rare" ]
```

Currently only has English translations, but feel free to comment below and I'll add the others in.


tests/filters/test_monster_filter.py
```diff
@@ -41,6 +41,27 @@ def test_monster_id(self):
...
+
+        # Generate events that should fail
+        fail1 = Events.MonEvent(generate_monster({"rarity": 1}))
+        fail2 = Events.MonEvent(generate_monster({"rarity": 2}))
+        fail3 = Events.MonEvent(generate_monster({"rarity": 4}))
+
+        # Test failing events
+        for e in [fail1, fail2, fail3]:
+            self.assertFalse(mon_filter.check_event(e))
+
```
billyjbryant: why not use the shortened method filter checks to reduce the space required and present a cleaner code appearance?

kvangent: Which filter checks are we talking about? This looks the same as the other tests to me

billyjbryant:
```python
     def test_cp(self):
         settings = {'min_cp': 20, 'max_cp': 500}
         mon_filter = Filters.MonFilter('cp_filter', settings)
         for i in [20, 250, 500]:
             self.assertTrue(mon_filter.check_event(create_event({'cp': i})))
 
         for i in [19, 501, 9999]:
             self.assertFalse(mon_filter.check_event(create_event({'cp': i})))
```

kvangent: Fixed in https://github.com/PokeAlarm/PokeAlarm/commit/9bfb0bcaadbd67c37fc20bacec8e187435e7a220.

PR #664, opened 2018-08-15
[WIP] Quest Events

evilmoses:
## Description
Ability to process quests sent to webhook.

## Type of Change

## Motivation and Context
Since scans are down nobody has worked on including quests in PA. My map now sends a quest webhook when a user submitted a new quest found so needed to process these in PA. Because these quests are not sent by scans, the information that we can send to the webhook is quite limited. Basically, the stop information and the text of the quest description and reward is sent. The expiry is dynamically calculated (23:59 tonight). Because it is just text that is sent, those requiring localization will need to ensure that the information is sent ALREADY TRANSLATED.

## How Has This Been Tested?
Has been tested and has been running on my prod environment for weeks, BUT......

My fork has not been updated to latest patch and is therefore behind the dev branch here. I have tested using latest for about 10 quests without any filters and it works, but needs serious further testing. Also, the following needs doing:

1. caching
2. filter testing (not tested at all)
3. unit tests
4. documentation

### DIFF
--- a/PokeAlarm/Alarms/Discord/DiscordAlarm.py
+++ b/PokeAlarm/Alarms/Discord/DiscordAlarm.py
@@ -81,11 +81,11 @@ class DiscordAlarm(Alarm):
             "url": "<gmaps>",
             "body": "The weather around <lat>,<lng> has changed to <weather>!"
         },
-        'quest': {
+        'quests': {
             'username': "Quest",
             'content': "",
-            'icon_url': None,
-            'avatar_url': None,
+            'icon_url': get_image_url("regular/quest/<type_id>.png"),
+            'avatar_url': get_image_url("regular/quest/<type_id>.png"),
             'title': "New Quest Found!",
             'url': "<gmaps>",
             'body': "Quest will expire at midnight."
@@ -122,8 +122,8 @@ class DiscordAlarm(Alarm):
             settings.pop('raids', {}), self._defaults['raids'])
         self.__weather = self.create_alert_settings(
             settings.pop('weather', {}), self._defaults['weather'])
-        self.__quest = self.create_alert_settings(
-            settings.pop('quest', {}), self._defaults['quest'])
+        self.__quests = self.create_alert_settings(
+            settings.pop('quests', {}), self._defaults['quests'])
 
         # Warn user about leftover parameters
         reject_leftover_parameters(settings, "'Alarm level in Discord alarm.")
@@ -230,7 +230,7 @@ class DiscordAlarm(Alarm):
 
     def quest_alert(self, quest_info):
         self._log.debug("Quest notification triggered.")
-        self.send_alert(self.__quest, quest_info)
+        self.send_alert(self.__quests, quest_info)
 
     # Send a payload to the webhook url
     def send_webhook(self, url, payload):
--- a/PokeAlarm/Alarms/Discord/DiscordAlarm.py
+++ b/PokeAlarm/Alarms/Discord/DiscordAlarm.py
@@ -88,7 +88,7 @@ class DiscordAlarm(Alarm):
             'avatar_url': get_image_url("regular/quest/<type_id>.png"),
             'title': "New Quest Found!",
             'url': "<gmaps>",
-            'body': "Quest will expire at midnight."
+            'body': "Quest will expire in <time_remaining>"
         }
     }
 
--- a/PokeAlarm/Alarms/Discord/DiscordAlarm.py
+++ b/PokeAlarm/Alarms/Discord/DiscordAlarm.py
@@ -80,6 +80,15 @@ class DiscordAlarm(Alarm):
             "title": "The weather has changed!",
             "url": "<gmaps>",
             "body": "The weather around <lat>,<lng> has changed to <weather>!"
+        },
+        'quest': {
+            'username': "Quest",
+            'content': "",
+            'icon_url': None,
+            'avatar_url': None,
+            'title': "New Quest Found!",
+            'url': "<gmaps>",
+            'body': "Quest will expire at midnight."
         }
     }
 
@@ -113,6 +122,8 @@ class DiscordAlarm(Alarm):
             settings.pop('raids', {}), self._defaults['raids'])
         self.__weather = self.create_alert_settings(
             settings.pop('weather', {}), self._defaults['weather'])
+        self.__quest = self.create_alert_settings(
+            settings.pop('quest', {}), self._defaults['quest'])
 
         # Warn user about leftover parameters
         reject_leftover_parameters(settings, "'Alarm level in Discord alarm.")
@@ -205,14 +216,21 @@ class DiscordAlarm(Alarm):
 
     # Trigger an alert when a raid egg has spawned (UPCOMING raid event)
     def raid_egg_alert(self, raid_info):
+        self._log.debug("Raid Egg notification triggered.")
         self.send_alert(self.__eggs, raid_info)
 
     def raid_alert(self, raid_info):
+        self._log.debug("Raid notification triggered.")
         self.send_alert(self.__raids, raid_info)
 
     # Trigger an alert based on Weather info
     def weather_alert(self, weather_info):
+        self._log.debug("Weather notification triggered.")
         self.send_alert(self.__weather, weather_info)
+
+    def quest_alert(self, quest_info):
+        self._log.debug("Quest notification triggered.")
+        self.send_alert(self.__quest, quest_info)
 
     # Send a payload to the webhook url
     def send_webhook(self, url, payload):
--- a/PokeAlarm/Alarms/FacebookPage/FacebookPageAlarm.py
+++ b/PokeAlarm/Alarms/FacebookPage/FacebookPageAlarm.py
@@ -79,12 +79,11 @@ class FacebookPageAlarm(Alarm):
             'caption': None
         },
         'quests': {
-            'message': "*New quest for <reward>*\n"
-                       "<quest>",
+            'message': "*New quest for <reward>*",
             'image': get_image_url('regular/quest/<type_id>.png'),
             'link': '<gmaps>',
             'name': 'Quest',
-            'description': 'New quest for <reward>\n<quest>',
+            'description': '<quest>',
             'caption': None
         }
     }
--- a/PokeAlarm/Alarms/FacebookPage/FacebookPageAlarm.py
+++ b/PokeAlarm/Alarms/FacebookPage/FacebookPageAlarm.py
@@ -77,6 +77,15 @@ class FacebookPageAlarm(Alarm):
             'description': "The weather around <lat>,<lng>"
                            " has changed to <weather>!",
             'caption': None
+        },
+        'quests': {
+            'message': "*New quest for <reward>*\n"
+                       "<quest>",
+            'image': get_image_url('regular/quest/<type_id>.png'),
+            'link': '<gmaps>',
+            'name': 'Quest',
+            'description': 'New quest for <reward>\n<quest>',
+            'caption': None
         }
     }
 
@@ -106,6 +115,8 @@ class FacebookPageAlarm(Alarm):
             settings.pop('raids', {}), self._defaults['raids'])
         self.__weather = self.create_alert_settings(
             settings.pop('weather', {}), self._defaults['weather'])
+        self.__quests = self.create_alert_settings(
+            settings.pop('quests', {}), self._defaults['quests'])
 
         #  Warn user about leftover parameters
         reject_leftover_parameters(
@@ -178,6 +189,10 @@ class FacebookPageAlarm(Alarm):
     # Trigger an alert based on Weather info
     def weather_alert(self, weather_info):
         self.send_alert(self.__weather, weather_info)
+
+    # Quest Alert
+    def quest_alert(self, quest_info):
+        self.send_alert(self.__quests, quest_info)
 
     # Sends a wall post to Facebook
     def post_to_wall(self, message, attachment=None):
--- a/PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
+++ b/PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
@@ -52,10 +52,9 @@ class PushbulletAlarm(Alarm):
             'body': "The weather around <lat>,<lng> has changed to <weather>!"
         },
         'quests': {
-            'message': "*New quest for <reward>*\n"
-                       "<quest>",
+            'title': '*New quest for <reward>*',
             'url': '<gmaps>',
-            'body': 'New quest for <reward>\n<quest>'
+            'body': '<quest>'
         }
     }
 
--- a/PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
+++ b/PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
@@ -50,6 +50,12 @@ class PushbulletAlarm(Alarm):
             'title': "The weather has changed!",
             'url': "<gmaps>",
             'body': "The weather around <lat>,<lng> has changed to <weather>!"
+        },
+        'quests': {
+            'message': "*New quest for <reward>*\n"
+                       "<quest>",
+            'url': '<gmaps>',
+            'body': 'New quest for <reward>\n<quest>'
         }
     }
 
@@ -81,6 +87,8 @@ class PushbulletAlarm(Alarm):
             settings.pop('raids', {}), self._defaults['raids'])
         self.__weather = self.create_alert_settings(
             settings.pop('weather', {}), self._defaults['weather'])
+        self.__quest = self.create_alert_settings(
+            settings.pop('quests', {}), self._defaults['quests'])
 
         #  Warn user about leftover parameters
         reject_leftover_parameters(
@@ -156,6 +164,10 @@ class PushbulletAlarm(Alarm):
     # Trigger an alert based on Weather info
     def weather_alert(self, weather_info):
         self.send_alert(self.__weather, weather_info)
+
+    # Trigger quest alert
+    def quest_alert(self, quest_info):
+        self.send_alert(self.__quest, quest_info)
 
     # Attempt to get the channel, otherwise default to all devices
     def get_sender(self, channel_tag):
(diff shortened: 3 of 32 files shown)


PokeAlarm/Filters/QuestFilter.py
```diff
@@ -0,0 +1,79 @@
...
+        self.max_dist = self.evaluate_attribute(  # f.max_dist <= m.distance
+            event_attribute='distance', eval_func=operator.ge,
+            limit=BaseFilter.parse_as_type(float, 'max_dist', data))
+            
+        # Quest Description
+        self.quest_contains = self.evaluate_attribute(  # f.gn matches e.gn
+            event_attribute='quest', eval_func=GymUtils.match_regex_dict,
+            limit=BaseFilter.parse_as_set(
+                GymUtils.create_regex, 'quest_contains', data))
+        self.gym_name_excludes = self.evaluate_attribute(  # f.gn no-match e.gn
```
LegitDongo: I thought quests only came from pokestops?
Why does this reference gyms instead of stops?
Glad you're adding filtering by stop name though 👍

LegitDongo (commented):
I haven't done any testing yet, just throwing out what I've seen off-hand.

PokeAlarm/Filters/QuestFilter.py
evilmoses: quests do only come from stops. It only references to contains function in the gym_utils as didn't find it anywhere else, lazy I know.

I haven't added filtering by stop name?? Only quest description and reward description.

The filtering needs more work, I don't use it (not enough user submissions to bother), and just added a couple of filters last minute for those that might need it - no idea if they even work....

I did this in about 30 minutes to get quest reporting working on my server. Needs more work I know, but people kept asking me for it, so easier to PR and let people fill in the blanks.

LegitDongo: I should have read a little more, and I need to start drinking coffee, my bad -- `'quest_contains'` 😅 
I wasn't expecting it to work the same way as it would by stops, but it makes sense now that I think about it
We will add more filters later once we actually have it merged, so let's focus on getting this ship-shape.
Something we should consider (maybe in the future if @kvangent doesn't deem it necessary now) is moving create_regex (as well as the other regex functions for good measure) into the GenUtils file since it will be used for more than just gyms, and to eliminate my confusion
Likewise, we will probably want to add some specific names for quests, since using `gym_name_excludes` doesn't really fit the bill - maybe make something more generic? such as a `contains` and `excludes` function somewhere

evilmoses: Oops, gym_name_excludes shouldn't be there lol

evilmoses: Gone.

nileplumb:
Can't wait this will hopefully encourage the few map forks that support report quest into some sort of order.
Could you also please document everything to the readthedocs / wiki too? Thanks!

kvangent:
Does the quest reward change? If it's sent with each webhook why are we trying to cache it?

evilmoses:
@kvangent, reason for caching is to prevent duplication. Because it is user submitted data, I am finding multiple users submitting the same quest by accident, so want to prevent duplicate posts making it to discord. Just need to check that the same quest/reward has not been sent on same day already etc.

PokeAlarm/Manager.py
```diff
@@ -946,6 +979,56 @@ def process_weather(self, weather):
...
+
+        # Calculate distance and direction
+        if self.__location is not None:
+            quest.distance = get_earth_dist(
+                [quest.lat, quest.lng], self.__location, self.__units)
+            quest.direction = get_cardinal_dir(
+                [quest.lat, quest.lng], self.__location)
+
+        # Store copy of cache info
+        # TODO
```
kvangent: I think you'll want something like the following: 

```
previous_reward = self._cache.quest_reward(quest.stop_id)
if quest.reward == previous_reward:
    self._log.debug("Quest ignored: Reward previously alerted!")
    return

self._cache.quest_reward(stop_id, quest.reward)
```

But it's worth noting it the Reward is a string, then it'll probably be wrong a lot. If possible we should try to represent the Reward as an enum.

PokeAlarm/Events/QuestEvent.py
```diff
@@ -0,0 +1,66 @@
...
+class QuestEvent(BaseEvent):
+    """ Event representing the discovery of a PokeStop. """
+
+    def __init__(self, data):
+        """ Creates a new Quest Event based on the given dict. """
+        super(QuestEvent, self).__init__('quest')
+
+        # Identification
+        self.stop_id = data['pokestop_id']
+        self.stop_name = data['name']
```
LegitDongo: It seems a bit ambiguous to have `name` come from the webhook when `name` can also be interpreted as `quest`.
I suggest renaming to `pokestop_name` to match `pokestop_id`.

LegitDongo (commented):
Caught this

PokeAlarm/Alarms/Discord/DiscordAlarm.py
```diff
@@ -80,6 +80,15 @@ class DiscordAlarm(Alarm):
             "title": "The weather has changed!",
             "url": "<gmaps>",
             "body": "The weather around <lat>,<lng> has changed to <weather>!"
+        },
+        'quest': {
```
LegitDongo: We should stick with the plural version -- `quests` -- since everything else uses the plural version

PokeAlarm/Alarms/Discord/DiscordAlarm.py
```diff
@@ -113,6 +122,8 @@ def __init__(self, mgr, settings, max_attempts, static_map_key):
             settings.pop('raids', {}), self._defaults['raids'])
         self.__weather = self.create_alert_settings(
             settings.pop('weather', {}), self._defaults['weather'])
+        self.__quest = self.create_alert_settings(
+            settings.pop('quest', {}), self._defaults['quest'])
```
LegitDongo: Same as above

LegitDongo (commented):
Specified for DiscordAlarm but should apply for the other types

PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
```diff
@@ -50,6 +50,12 @@ class PushbulletAlarm(Alarm):
             'title': "The weather has changed!",
             'url': "<gmaps>",
             'body': "The weather around <lat>,<lng> has changed to <weather>!"
+        },
+        'quests': {
+            'message': "*New quest for <reward>*\n"
```
kvangent: s/message/title

title -> `"*New quest for  <reward>*"`
body -> `"<quest>"`

PokeAlarm/Alarms/FacebookPage/FacebookPageAlarm.py
```diff
@@ -77,6 +77,15 @@ class FacebookPageAlarm(Alarm):
             'description': "The weather around <lat>,<lng>"
                            " has changed to <weather>!",
             'caption': None
+        },
+        'quests': {
+            'message': "*New quest for <reward>*\n"
```
kvangent: To fit in with the other events in this alarm, split message/description on the new line and put the first half in message and second half in description.

PokeAlarm/Alarms/Discord/DiscordAlarm.py
```diff
@@ -80,6 +80,15 @@ class DiscordAlarm(Alarm):
...
             "body": "The weather around <lat>,<lng> has changed to <weather>!"
+        },
+        'quests': {
+            'username': "Quest",
+            'content': "",
+            'icon_url': get_image_url("regular/quest/<type_id>.png"),
+            'avatar_url': get_image_url("regular/quest/<type_id>.png"),
+            'title': "New Quest Found!",
+            'url': "<gmaps>",
+            'body': "Quest will expire at midnight."
```
kvangent: Is there a time remaining DTS we could use here?

PokeAlarm/Events/QuestEvent.py
```diff
@@ -0,0 +1,76 @@
...
+# 3rd Party Imports
+# Local Imports
+from PokeAlarm import Unknown
+from . import BaseEvent
+from PokeAlarm.Utils import get_gmaps_link, get_applemaps_link, \
+    get_waze_link, get_dist_as_str
+
+
+class QuestEvent(BaseEvent):
+    """ Event representing the discovery of a PokeStop. """
```
kvangent: Update

PokeAlarm/Events/QuestEvent.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # Used to reject
+        self.name = self.stop_id
+        self.geofence = Unknown.REGULAR
+        self.custom_dts = {}
+
+        # Quest Details
+        self.quest = data['quest']
+        self.reward = data['reward']
+        self.expiry = datetime.datetime.now().strftime("%d/%m/%Y 23:59")
```
kvangent: Does this field not get sent in the webhook? Might cause issues if processing is delayed (or sent around midnight or something) 

Also `expire` or `expire_time` would probably be better names for the field.

PokeAlarm/Events/QuestEvent.py
```diff
@@ -0,0 +1,76 @@
...
+        self.type = check_for_none(int, data.get('type'), 0)
```
kvangent: Might be some confusion with mon type - maybe `reward_type`?

PokeAlarm/Manager.py
kvangent: This is still TODO?

PokeAlarm/Utilities/QuestUtils.py
```diff
@@ -0,0 +1,47 @@
...
+                    j = j['quest_types']
+                    for id_ in j:
+                        nm = j[id_].lower()
+                        get_type.ids[nm] = int(id_)
+        if name in get_type.ids:
+            return get_type.ids[name]
+        else:
+            return int(name)  # try as an integer
+    except ValueError:
+        raise ValueError("Unable to interpret `{}` as a valid "
```
kvangent, line 28: This error message needs to be updated.

PokeAlarm/Utilities/QuestUtils.py
```diff
@@ -0,0 +1,47 @@
...
+                         " severity name or id.".format(reward_type))
+    # if isinstance(type_id, int):
+    #     try:
+    #         if not hasattr(get_type, 'types'):
+    #             get_type.types = {}
+    #             files = glob(get_path('locales/*.json'))
+    #             for file_ in files:
+    #                 with open(file_, 'r') as f:
+    #                     j = json.loads(f.read())
+    #                     j = j['quest_types']
```
kvangent: Does this section still need to be here?

locales/zh_hk.json
```diff
@@ -948,6 +948,16 @@
...
   },
+  "quest_types": {
+    "000": "Unset",
+    "001": "Experience",
+    "002": "Item",
+    "003": "Stardust",
+    "004": "Candy",
+    "005": "Avatar Clothing",
+    "006": "Quest",
+    "007": "Monster Encounter"
```
kvangent: Probably should at least Google Translate these until people complain

prototype/events/quest.py
```diff
@@ -0,0 +1,43 @@
...
+    # Identification
+    stop_id = EventAttr(['pokestop_id'], str, required=True)  # type: str
+    type = EventAttr(['type'], int, default=0)  # type: int
+
+    # Details
+    quest = EventAttr(['quest'], str, required=True)  # type: str
+    reward = EventAttr(['reward'], str, required=True)  # type: str
+
+    @property
+    def expiry(self):
```
kvangent: This property needs a better name.

prototype/events/quest.py
```diff
@@ -0,0 +1,43 @@
...
+class Quest(Event):
+    """
+    Object representation of a change to a quest.
+    """
+
+    # Location
+    lat = EventAttr(['latitude'], float)  # type: float
+    lng = EventAttr(['longitude'], float)  # type: float
+    # Completed by Manager
+    distance = None  # type: float
```
kvangent, line 27: Mark these with TODO please

start_pokealarm.py
```diff
@@ -14,7 +14,7 @@
 import sys
 # 3rd Party Imports
 import configargparse
-from gevent import wsgi, spawn, signal, pool, queue
+from gevent import pywsgi, spawn, signal, pool, queue
```
kvangent, line 17: Why this change?

kvangent (changes requested):
Also need wiki updates, but they can be in a separate PR if desired

PokeAlarm/Events/QuestEvent.py
LegitDongo: It's not sent in the webhook
Quests always expire at midnight so there's not much point to send it.

start_pokealarm.py
LegitDongo: gevent==1.2.2 is what we have in master, and 1.3.4 is in dev
As of gevent 1.3.0, wsgi was renamed to pywsgi
I probably should have just put this in a seperate PR 😅

PokeAlarm/Events/QuestEvent.py
kvangent: Without sending the date, there is the possibility of introducing a bug where a webhook gets sent before midnight, is delayed for some reason, and isn't received until after. When PA displays the info it would then be incorrect. It's probably not a huge priority, but something to consider.

prototype/events/quest.py
LegitDongo: How about `expire_time`?

kvangent: Sounds good to me.

PokeAlarm/Events/QuestEvent.py
fosJoddie: Research Tasks change either at midnight or when Niantic starts/ends events that also changes the research task pool. 

For community days, this means research tasks will change 3 times during the day: midnight, 11 local time and finally back to the originals (src needed) at 14 local time.

For days that end or start events (like Halloween event) this means tasks rotate some arbitary time of the day, which you cannot easily know, since event times are special.

I don't see the point of implementing an expire_time at all for any purpose. The only reason why you'd want it is if PA queue is large and you somehow want PA to ignore too old data. This could be better done with a `datetime.datetime.now()` and check if it's still the same day.

PokeAlarm/Manager.py
fosJoddie: The check has to check if both reward AND task (quest) is the same. Since two different tasks might give same reward.

PokeAlarm/Cache/Cache.py
```diff
@@ -97,6 +98,12 @@ def day_or_night_id(self, s2_cell_id, day_or_night_id=Unknown.REGULAR):
             self._day_or_night_id[s2_cell_id] = day_or_night_id
         return self._day_or_night_id.get(s2_cell_id, Unknown.REGULAR)
 
+    def quest_reward(self, stop_id, reward=None):
```
fosJoddie: the uniqueness of quests is not only the stop_id + reward. There's also the quest/task itself.

the Cleanup method also needs some way to flush out all old rewards at midnight, or else pokestops that happen to have the same task/reward two days in a row will never get reported.

fosJoddie (changes requested):

PokeAlarm/Events/QuestEvent.py
kvangent: I still think this field should be sourced from the scanner side. I disagree that it doesn't have purpose - the field seems to have value for deduping, resolving the issues with delays, and with areas around horizons of time change. Presumably it's somewhat relevant for the user to know how long they have to pick up the quest as well?

LegitDongo: I can see how this is an issue on both sides, but using `datetime.datetime.now()` would force an expire time when it's possible the owner of the manual submissions may be aware of the varying expire times for quests on days such as community days. Likewise, there probably won't be much of a queue at to really have a need for an expire time field.

From where I'm sitting there's no easy way to solve both issues without adding another webhook field -- `override` or something like that -- which will tell PA to just ignore the cache.

fosJoddie: The protos in the real game suggests that they use an "updated" timestamp to figure out if a research task is the same or not. 
If the field is a "was discovered at TT" or "expected to expire at TT" doesn't make much difference to me, I'm just waffling out my own thoughts here.  Either way, if this field stays it will be more work for any tools wanting to send data to PA

PokeAlarm/Cache/FileCache.py
```diff
@@ -42,6 +42,7 @@ def _load(self):
                 self._cell_weather_id = data.get('cell_weather_id', {})
                 self._severity_id = data.get('severity_id', {})
                 self._day_or_night_id = data.get('day_or_night_id', {})
+                self._reward = data.get('reward', {})
```
kvangent, line 45: Looks like `_task` is missing?

PokeAlarm/Cache/FileCache.py
```diff
@@ -64,7 +65,8 @@ def _save(self):
             'gym_image': self._gym_image,
             'cell_weather_id': self._cell_weather_id,
             'severity_id': self._severity_id,
-            'day_or_night_id': self._day_or_night_id
+            'day_or_night_id': self._day_or_night_id,
+            'reward': self._reward
```
kvangent: `_task` here too?

PokeAlarm/Events/QuestEvent.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # Used to reject
+        self.name = self.stop_id
+        self.geofence = Unknown.REGULAR
+        self.custom_dts = {}
+
+        # Quest Details
+        self.quest = data['quest']
+        self.reward = data['reward']
+        self.expire_time = datetime.datetime.now().strftime("%d/%m/%Y 23:59")
```
kvangent: Where did we land on this? Seems like taking an argument with a default would be best?

kvangent (changes requested):

kvangent (approved):
LGTM

PR #670, opened 2018-10-26
Support for Ex Raid Eligibility

kvangent:
Added 'ex_eligible' parameter from the webhook field 'is_ex_raid_eligible' to Gym, Egg, and Raid events. Accepts a boolean value. 

Added 'is_ex_eligible' to Gym, Egg, and Raid filters. Accepts a boolean value.


LegitDongo (dismissed):
I see no obvious issues with this, other than this not being sent in gym webhooks that I'm aware of.

But that's not to say it won't be later, since it's very likely.

PR #672, opened 2018-11-23
Quests

Canucksfan24:
## Type of Change

## Wiki Update


LegitDongo:
This was probably an accident since it's just commits made on the quests PR.

PR #673, opened 2018-11-25
Update prototype to Python3 compatibility only. 

kvangent:
WIP. Open to verify testing config for Travis.


pokealarmv4/events/_base.py
```diff
@@ -0,0 +1,61 @@
...
+
+This module contains base classes for the creation of the Events module.
+"""
+
+from typing import Any, Callable, List, Mapping, NamedTuple, Optional
+
+
+class EventAttr(NamedTuple):
+    """
+    EvenAttr is a helper class for define EventAttr. When defined in an Event
```
LegitDongo, line 16: Circular definition, and misspelling

pokealarmv4/events/raid.py
```diff
@@ -0,0 +1,55 @@
...
+from ._base import Event, EventAttr
+
+
+class Raid(Event):
+    """
+    Object representation of an ingame Raid.
+    """
+
+    # Identification
+    gym_id = EventAttr(['gym_id', 'id'], str, required=True)
```
LegitDongo, line 21: no `str` type hint

LegitDongo (changes requested):
Other than some minor changes I have a few concerns:
Where is all the information going that was in `data`?
Are we planning on having multiple versions of PA in the same project?
Are we going to allow for multiple Python versions?

The new format looks spectacular btw, but I'm not going to pretend like I know everything that's going on

PR #677, opened 2019-02-03
Time DTS without seconds

LegitDongo:
## Description
Time DTS without seconds
In the case of hours being available, it will show hours as well as minutes
Includes documentation update

This was a frequently asked for feature

## Type of Change

## Motivation and Context
Frequently requested.
Some people don't like the seconds on the end
It also makes PA compatible with other tools

## How Has This Been Tested?
Locally, Windows 10

## Wiki Update

### DIFF
--- a/PokeAlarm/Events/EggEvent.py
+++ b/PokeAlarm/Events/EggEvent.py
@@ -79,6 +79,14 @@ class EggEvent(BaseEvent):
             '12h_raid_end': raid_end_time[1],
             '24h_raid_end': raid_end_time[2],
 
+            # Time Remaining Without Seconds
+            'hatch_time_left_minutes': hatch_time[3],
+            '12h_hatch_time_minutes': hatch_time[4],
+            '24h_hatch_time_minutes': hatch_time[5],
+            'raid_time_left_minutes': raid_end_time[3],
+            '12h_raid_end_minutes': raid_end_time[4],
+            '24h_raid_end_minutes': raid_end_time[5],
+
             # Location
             'lat': self.lat,
             'lng': self.lng,
--- a/PokeAlarm/Events/EggEvent.py
+++ b/PokeAlarm/Events/EggEvent.py
@@ -71,21 +71,27 @@ class EggEvent(BaseEvent):
             # Identification
             'gym_id': self.gym_id,
 
-            # Time Remaining
+            # Hatch Time Remaining
             'hatch_time_left': hatch_time[0],
             '12h_hatch_time': hatch_time[1],
             '24h_hatch_time': hatch_time[2],
+            'hatch_time_no_secs': hatch_time[3],
+            '12h_hatch_time_no_secs': hatch_time[4],
+            '24h_hatch_time_no_secs': hatch_time[5],
+            'hatch_time_raw_hours': hatch_time[6],
+            'hatch_time_raw_minutes': hatch_time[7],
+            'hatch_time_raw_seconds': hatch_time[8],
+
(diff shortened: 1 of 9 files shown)


PokeAlarm/Utils.py
```diff
@@ -481,7 +481,15 @@ def get_time_as_str(t, timezone=None):
...
     time_24 = disappear_time.strftime("%H:%M:%S")
-    return time_left, time_12, time_24
+
+    # Get the same as above but without seconds
+    time_left_minutes = "%dm" % m if h == 0 else "%dh %dm" % (h, m)
+    time_12_minutes = disappear_time.strftime("%I:%M") \
+        + disappear_time.strftime("%p").lower()
+    time_24_minutes = disappear_time.strftime("%H:%M")
+
+    return time_left, time_12, time_24, \
```
kvangent:
```suggestion
    return time, time_12h, time_24h, \
```

PokeAlarm/Utils.py
```diff
@@ -481,7 +481,15 @@ def get_time_as_str(t, timezone=None):
...
+        time_left_minutes, time_12_minutes, time_24_minutes
```
kvangent: v
```suggestion
        time_no_sec, time_12h_no_sec, time_24h_no_sec
```

docs/configuration/events/egg-events.rst
```diff
@@ -75,16 +75,22 @@ geofence            Geofence around the event.
...
+======================= ===============================================================
+DTS                     Description
+======================= ===============================================================
+hatch_time_left         Time remaining until the egg hatches.
+12h_hatch_time          Time when the egg will hatch, formatted in 12h.
+24h_hatch_time          Time when the egg will hatch, formatted in 24h.
+raid_time_left          Time remaining until the raid ends.
+12h_raid_end            Time when the raid ends, formatted in 12h.
+24h_raid_end            Time when the raid ends, formatted in 24h.
+hatch_time_left_minutes Time remaining until the egg hatches without seconds.
```
kvangent: Maybe we should update these more explicitly? (formatted h?:mm:ss) vs (formatting h?:mm)

PokeAlarm/Events/EggEvent.py
```diff
@@ -79,6 +79,14 @@ def generate_dts(self, locale, timezone, units):
             '12h_raid_end': raid_end_time[1],
             '24h_raid_end': raid_end_time[2],
 
+            # Time Remaining Without Seconds
```
kvangent: Separating on seconds/no_seconds seems weird, maybe separate on hatch_time, raid_end_time instead?

PokeAlarm/Events/EggEvent.py
```diff
@@ -79,6 +79,14 @@ def generate_dts(self, locale, timezone, units):
...
+            'hatch_time_left_minutes': hatch_time[3],
```
kvangent: I don't like minutes, because it implies that the other doesn't have minutes. How about `hatch_time_no_secs`?

kvangent (changes requested):

kvangent:
@LegitDongo Can we address feedback and get this merged?

docs/configuration/events/egg-events.rst
LegitDongo: I'm thinking of adding an example column so that it's defined a little more explicitly

kvangent (dismissed):

kvangent (approved):

PR #678, opened 2019-02-21
Separate caching for encountered and non encountered mon

KartulUdus:
## Description

Current scanners send non-encountered and encountered mons separately with the same encounter_id

this stops users from receiving the alarm when it is encountered. 

I've changed the cache key to be `encounter_id + weight` so that once encountered and the monsters weight exists, it will be a different cache entry.

I also have disappear time in there in case some manual submission or something sends the same encounter_id, but I didn't know if you would see value in that so left it out for now.

## Type of Change

## Motivation and Context

https://discordapp.com/channels/215181169761714177/218822834225545216/547635147621531648
## How Has This Been Tested?

None, I did this in an accounting meeting
## Wiki Update


LegitDongo:
I don't have an issue with this, but this isn't how we were originally going to go about this.
I think @kvangent should have something to say on this.

I plan on PRing to fix new Travis issues, so prepare for a rebase in the near future 😄

kvangent:
I'm good with this - it's not as complete of a solution as we have planned, but given my progress on v4 is pretty slow this seems like a reasonable workaround.

nhumber:
im getting this error using MAD

Encountered error during processing: TypeError: unsupported operand type(s) for +: 'long' and 'str'
2019-02-24 17:34:59,883 [ERROR][ pokealarm][ manager_0] Stack trace:
  File "/home/user/PokeAlarm/PokeAlarm/Manager.py", line 520, in run
    self.process_monster(event)
  File "/home/user/PokeAlarm/PokeAlarm/Manager.py", line 626, in process_monster
    if self.__cache.monster_expiration(mon.enc_id + str(mon.weight)) is not None:
TypeError: unsupported operand type(s) for +: 'long' and 'str'

KartulUdus:
interesting @nhumber 
seems that MAD is sending encounter as "long" not a "str" type

changed it to convert both to string, could you try again?

nhumber:
thanks, error free and functioning.

kvangent:
@LegitDongo Can you please merge this this weekend?

LegitDongo (approved):
Gave a lightweight test, seems to work appropriately.

PR #684, opened 2019-05-27
New enhancements for pokestop event

JimBim89:
## Description
Adds name, image and lure type to pokestop event.
Also added incident/invasion support including types.

## Type of Change

## Motivation and Context
This PR adds support for new special lures and adds additional info.
Now also includes incident/invasions with grunt type information.

## How Has This Been Tested?
This has been tested with MAD using PRs I submitted.
Also in contact with some people for RDM compatibility.


JoniValles:
Maybe it's a good idea to add a filter by 'lure_type_name'.

LegitDongo (changes requested):
Sorry for the delay, been away for a bit

Is `active_fort_modifier` how people are sending this field? This seems like an odd naming choice to me.

Other than this, we should probably add naming for the other locales as well. A simple Google Translate will suffice for now.

Also, totally agree with adding the lure type filter. This would be best to accept an array - since most people will want to filter by only the three new types.

JimBim89:
Well I thought about the naming myself but stopped once I found it here https://pa.readthedocs.io/en/master/miscellaneous/webhook-standard.html#pokestops

And it's the original name from protos.

LegitDongo:
I'm pretty sure that field is used for something else

I've been meaning to ask Flo how they send this field

I'll post back when I've got a response.

LegitDongo:
Flo said that they are sending it as `lure_id`, so that should get changed as well.

sn0opy:
> Flo said that they are sending it as `lure_id`, so that should get changed as well.

This has just been adapted by MAD https://github.com/Map-A-Droid/MAD/commit/fed3ac487caa310a4308eb502eb832dfdd2dfacd

LegitDongo:
A filter for this is still needed.
Do you need help with this?

JimBim89:
Updated, any feedback is welcome! If you do a manual merge make sure to get all the locale files, otherwise the filter types will error out on missing array items. If incidents come back, maybe try the new caching mechanism by luring an already invaded stop.

LegitDongo:
Relaying what was said in Discord for reference:
`last_modified` gets changed whenever it's been lured or been invaded, so that should be used for cache checking
The alternative is to check both the lure id and the invasion info, thus why I think `last_modified` would be better

Also, some filters/DTS need added for invasions.

I also am considering making a new icon for these, just not sure what to go with or how to implement it. (suggestions from anyone here would be helpful)

RDM adds a new webhook type -- `invasion` -- which is basically a stop with a few fields added.

There seems to be some consistency between the reward monster and the rocket member shown, but this will have to be added later due to the game changes lately.

If you can't get to any of these extras I'll be available to start on this stuff sometime after next week.

LegitDongo:
Reference for RDM invasion type
```
   } else if type == "invasion" {
            let message: [String: Any] = [
                "pokestop_id":id,
                "latitude":lat,
                "longitude":lon,
                "name":name ?? "Unknown",
                "url":url ?? "",
                "lure_expiration":lureExpireTimestamp ?? 0,
                "last_modified":lastModifiedTimestamp ?? 0,
                "enabled":enabled ?? true,
                "lure_id":lureId ?? 0,
                "pokestop_display":pokestopDisplay ?? 0,
                "incident_expire_timestamp":incidentExpireTimestamp ?? 0,
                "updated": updated ?? 1
            ]
            return [
                "type": "invasion",
                "message": message
            ]
```

LegitDongo:
After consulting the MAD devs, it appears that a that if a worker returns and a stop gets observed as invaded and lured at the same time, it will be sent in one webhook. This causes some issues since people will expect two webhooks to come out of this; one for invasions and one for lures.

This considered, it might be wise to make a new event type for invasions so that it's easier to handle between RDM and MAD, and also make it easier to fire off multiple events in the situation listed above.

Also, `last_modified` should *not* be the way to go for this reason.

JimBim89:
Updated with grunt_type support and RDM compatibility. 

Lots of untested code, so let me know if everything still works.

LegitDongo:
A few things to consider:
We should probably take into account the gender of the grunt as well in order to provide more accurate images whenever we add them for this, making a nicer variety.
Also, we should include a filter for the following:
1. Type
2. Tier

"Tier 2" guarantees a mon from this [list](https://cdn.discordapp.com/attachments/546982390413787136/605754702004486189/image0.png)
Although we don't know the typing for tier 2 mons, we at least know that it will be worth going for.
See [here](https://cdn.discordapp.com/attachments/546982390413787136/605754436853432350/image0.png) for example

We can keep the current grunt_ids filters, since some may prefer to do it this way.

Also, there's a possibility of either lures or invasions getting missed with MAD, since they are both sent in the same webhook and could be observed by the worker at the same time, thus sent together. In these cases, two events should be fired, which might get a little hairy.
That being said, I'm working on getting something worked out for this, with hopes that the MAD team will adopt it.

0xC1A551C:
Why is this not merged yet?

LegitDongo:
@0xC1A551C There is some discrepancy on 
1. How it gets sent
2. If we choose to go the MAD route, missing either a lure or a invasion since you can only get one alert per webhook currently.

We also need a few filters/DTS added.

I haven't had much time to work on it.

LegitDongo:
Alright guys, I got a decent implementation going now
Some notes
# !!This will break your current configuration!!
Do ***not*** pull unless you are ready for this
See [here](https://github.com/PokeAlarm/PokeAlarm/pull/684/files#diff-b06b4ad238893b51ae77e7e4311cb1bf) for filter listings
Current invasion type DTS:
`stop_id`, `stop_name`, `stop_image`, `type_id`, `type_name`, `gender_id`, `gender`, standard time DTS, standard lat/lng/distance, etc

### New Features 
1. `invasion` is now a type to be put in filters/alarms
2. This new set will allow both a lure and invasion webhook to be sent when sent in the same webhook
3. New grunt types added

Images and additional translations coming soon
This is ***very*** WIP, so tests would be greatly appreciated!

LegitDongo:
Added `type_emoji` for DTS
Updated all locales files to reflect monster types and Google Translated everything else
Added more docs

kvangent (dismissed):
LGTM

kvangent (approved):

LegitDongo (approved):

PR #685, opened 2019-06-17
Add gender field support for raid event

JimBim89:
## Description

Adds gender information to raid event.

## Type of Change

## Motivation and Context

This PR adds support for gender (useful for Burmy, Snorunt etc).

## How Has This Been Tested?

This has been tested with multiple webhooks and should work.


LegitDongo:
Adding some filters would also be a good idea, like we already have for monster events
[gender monster filter](https://github.com/PokeAlarm/PokeAlarm/blob/master/PokeAlarm/Filters/MonFilter.py#L114)

LegitDongo (approved):
Got a test from Duke
Looks good
Thanks for the PR!

PR #686, opened 2019-07-12
Added missing moves to en.json and move_info.json

mrcbat:
## Description

Added the missing moves to the move_info.json file and to en.json

## Type of Change

- [ x] Bug fix (non-breaking change which fixes an issue)

## Motivation and Context

Currently the moves show as Unknown in PokeAlarm, this fixes that (at least in english locales)

## How Has This Been Tested?

Tested in my local Discord

## Wiki Update

- [x ] This change does not require an update to the Wiki.


LegitDongo:
Closing in favor of #688 since that has damage, dps, duration, and energy filled out

JimBim89:
Would agree but 0 is better than copy pasted values which don't reflect correct dts?

PR #692, opened 2019-08-07
Fix Geofence Naming in example.json

nhumber:
## Type of Change

## Wiki Update


LegitDongo:
This was already fixed in dev.
It will get merged into master eventually

Issue #691, opened 2019-08-06
Python shenanigans (data.get or compatibility)


JabLuszko:
## Troubleshooting Checklist

- [x] My Scanner install is at the most recent version.
- [x] My PokeAlarm install is at the most recent version.
- [x] I've already checked the
 [FAQ](https://github.com/RocketMap/PokeAlarm/wiki/faq) and the
 [Wiki](https://github.com/RocketMap/PokeAlarm/wiki) already.

## Problem Description
This was debugged inside WeatherEvent, but should impact more things that use diff fields for MAD/RDM.
Initial error was: ```ValueError: '=' alignment not allowed in string format specifier```
Turns out severity_id was None, while it was sent in webhook as "alert_severity" - yet "severity" (probably used by RDM) was not set, so was returning None.

```
self.severity_id = data.get('alert_severity') or data.get('severity')

Log:
data.get('alert_severity'): 0
data.get('severity'): None

>>> print 0 or None
None
>>> print None or 0
0

Python3 does the same thing.
```

## Environment
Operating System: ` Linux `
Python Version: ` Python 2.7.15+ (default, Nov 27 2018, 23:36:35)
[GCC 7.3.0] on linux2`

LegitDongo:
`alert_severity` was getting sent as `0` - a valid value
I'm assuming that Python evaluates this as "falsey", thus the `or` skips over it

LegitDongo:
https://docs.python.org/2/reference/expressions.html#boolean-operations
Documentation reference
```
The expression x and y first evaluates x; if x is false, its value is returned; otherwise, y is evaluated and the resulting value is returned.

The expression x or y first evaluates x; if x is true, its value is returned; otherwise, y is evaluated and the resulting value is returned.
```

JimBim89:
Line 67 may also be a culprit here, changing "is" to "==" should be the way to go.

(WeatherEvent.py:67:23: F632 use ==/!= to compare str, bytes, and int literals)

JimBim89:
Also when weather condition is 0 (ingame weather provider down) it will be displayed as unknown in PA alerts instead of using the available locale "None" for id 0.

LegitDongo:
The problem is that lazy evaluation treats `0` as `falsey`, so my plan was to change it to something like
`data.get('alert_severity', data.get('severity', 0))` in order to only go to the next `data.get` on `None`
e.g. 
[...]

PR #693, opened 2019-08-08
Fixed weather/severity event issues

JimBim89:
## Description
This fixes the issues with the or statement related to 0 ids.

## Type of Change

## Motivation and Context
Fixes for issue https://github.com/PokeAlarm/PokeAlarm/issues/691

## How Has This Been Tested?
Locally with latest dev.

## Wiki Update


JimBim89:
While digging through weather stuff I also noticed that PA has serverity "extreme" as 3, but in the protos it has a value of 2 ([link](https://github.com/Furtif/POGOProtos/blob/master/src/POGOProtos/Map/Weather/WeatherAlert.proto)), does someone know why its one off here? 

Also the text for "moderate" is never mentioned like that to the player in game, the weather alert is always displayed as "Careful", maybe adjust the locales to reflect the ingame behavior?

kvangent (dismissed):

kvangent:
> While digging through weather stuff I also noticed that PA has serverity "extreme" as 3, but in the protos it has a value of 2 ([link](https://github.com/Furtif/POGOProtos/blob/master/src/POGOProtos/Map/Weather/WeatherAlert.proto)), does someone know why its one off here?
> 
> Also the text for "moderate" is never mentioned like that to the player in game, the weather alert is always displayed as "Careful", maybe adjust the locales to reflect the ingame behavior?

The protos may have shifted, or we may have gotten it wrong in the first place. Happy to update if you want to PR it.

LegitDongo (dismissed):
This doesn't fix the underlying issue of `or` usage, however this fixes the immediate bug of weather being broken.

LegitDongo:
Waiting on merge until we can get a few more weather-related things put into this

LegitDongo:
@JimBim89 Here's the changes we discussed on Discord, did you want to add these here?
```
for severity it would error like mentioned avbove (can't shift unknown to 3 digits)
and for ingame weather disabled it wouldn't get the localized weather word for 0 (None)
whats still up for discussion is the id for extreme weather and the locals for moderate - in PA extreme is 3 but in protos its 2
and would you rather have the old "Moderate" text from protos var or the ingame text "Careful" for id 1?
```

We also discussed a better way to alert weather changes since severity can change several times a minute, such as a cooldown
`maybe a cooldown (1-5min?) after a severity change`
https://cdn.discordapp.com/attachments/218822834225545216/656562725882822722/Unbenannt.png

LegitDongo (approved):
Since this still fixes the bug I'm going to go ahead and merge this.

If you still want to add the mentioned functionality later you can open a new PR

PR #701, opened 2019-10-19
WIP - Add verified spawn emoji for DTS

DrMurx:
## Description

Adds a new DTS `spawn_verified_emoji` which translates the simple boolean of `spawn_verified` to  a ✅ or a ❓ emoji respectively.

## Type of Change

## Motivation and Context

Emojis are easier to grasp than a literal boolean.

## How Has This Been Tested?

By adding the new token to a DTS template

## Wiki Update

### DIFF
--- a/PokeAlarm/Events/MonEvent.py
+++ b/PokeAlarm/Events/MonEvent.py
@@ -143,25 +143,19 @@ class MonEvent(BaseEvent):
             'encounter_id': self.enc_id,
             'mon_name': locale.get_pokemon_name(self.monster_id),
             'mon_id': self.monster_id,
-            'mon_id_3': "{:03}".format(self.monster_id),
-
-            # Time Remaining
-            'time_left': time[0],
-            '12h_time': time[1],
-            '24h_time': time[2],
-            'time_left_no_secs': time[3],
-            '12h_time_no_secs': time[4],
-            '24h_time_no_secs': time[5],
-            'time_left_raw_hours': time[6],
-            'time_left_raw_minutes': time[7],
-            'time_left_raw_seconds': time[8],
-
+            'mon_id_3': "{:03}".format(self.monster_id)
+        })
+        dts.update(self._disappear_time_infos(time, self.spawn_verified))
+        dts.update({
             # Spawn Data
             'spawn_start': self.spawn_start,
             'spawn_end': self.spawn_end,
             'spawn_verified': self.spawn_verified,
             'spawn_verified_emoji': get_verified_spawn_emoji(
                 self.spawn_verified),
+            'unknown_spawn_end_or_empty': (
+                locale.get_unknown_despawn_time() if self.spawn_verified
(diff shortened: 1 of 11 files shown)


darkelement1987:
Could you please add your changes to the docs? And maybe use ❌ for 'Not verified'

DrMurx:
Sure. 

However, I'd like to develop the whole idea a bit further, will push some more commits to this PR tonight. Documentation will be updated as well.

JimBim89:
My local setup uses all three, using ? for no verified data.

I also have verified_emoji_or_empty and unverified_emoji_or_empty.

So you don't have to have checkpoints all the time.

PokeAlarm/Events/MonEvent.py
```diff
@@ -337,3 +333,20 @@ def generate_dts(self, locale, timezone, units):
                 and self.weight <= 2.41 else '')
         })
         return dts
+
+    def _disappear_time_infos(self, time, spawn_verified):
```
LegitDongo, line 337: This seems like something that would do better in the PokeAlarm.Utils file

PokeAlarm/Events/MonEvent.py
```diff
@@ -143,23 +143,19 @@ def generate_dts(self, locale, timezone, units):
...
-            'time_left': time[0],
-            '12h_time': time[1],
-            '24h_time': time[2],
-            'time_left_no_secs': time[3],
-            '12h_time_no_secs': time[4],
-            '24h_time_no_secs': time[5],
-            'time_left_raw_hours': time[6],
-            'time_left_raw_minutes': time[7],
-            'time_left_raw_seconds': time[8],
-
```
LegitDongo, line 158: Why were the above lines removed?
I think this is a standard part of dev now

LegitDongo (changes requested):
I like this, just a few questions

PokeAlarm/Events/MonEvent.py
DrMurx: I accidentally pushed a preliminary commit, that's why I marked the whole PR as WIP. It's seriously broken now, but I'm super-busy atm and will push the full changeset later.

DrMurx: As an outline, all the time fields get an `...or_empty` counterpart. Any given despawn time information is just an estimation if the despawn time is not verified, so the template configuration could reflect this.

PokeAlarm/Events/MonEvent.py
DrMurx: Sure :)

LegitDongo: Sorry, I meant PokeAlarm.MonUtils since this will only be used for monsters :smile:

LegitDongo:
@DrMurx what's the status of this?

LegitDongo:
Closed in favor of #710

PR #708, opened 2019-12-31
Custom Static Map Alarm Field

LegitDongo:
## Description
Allow the static map field to be through a custom URL for alarms

## Type of Change

## Motivation and Context
Google static maps are cool and all, but they can be expensive and some people may want more options.

## How Has This Been Tested?
very little, win10, webhook tester

## Wiki Update
Wiki update included.

(Ignore the first commit message, not sure how that happened)


temp666:
This works well with Discord. When is the plan to merge this?

kvangent (approved):

PR #710, opened 2020-01-22
Spawn Verified DTS

JimBim89:
## Description
This changes the verified data from a simple false/true to a unknown/no/yes.
It also improves the existing dts for this and adds additional emoji dts.

## Type of Change

## Motivation and Context
Prior to this you could not distinguish between no data at all and a unverified one.
And since nearly all my spawns are verified, I rather want to only display dts for unverified.

## How Has This Been Tested?
Tested with latest PokeAlarm and MAD.

## Wiki Update

### DIFF
--- a/PokeAlarm/Events/MonEvent.py
+++ b/PokeAlarm/Events/MonEvent.py
@@ -9,7 +9,7 @@ from PokeAlarm.Utils import (
     get_move_duration, get_move_energy, get_pokemon_size,
     get_applemaps_link, get_time_as_str, get_seconds_remaining,
     get_base_types, get_dist_as_str, get_weather_emoji,
-    get_type_emoji, get_waze_link)
+    get_verified_emoji, get_type_emoji, get_waze_link)
 from . import BaseEvent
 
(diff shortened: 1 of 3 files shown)


PokeAlarm/Events/MonEvent.py
```diff
@@ -159,7 +160,16 @@ def generate_dts(self, locale, timezone, units):
             # Spawn Data
             'spawn_start': self.spawn_start,
             'spawn_end': self.spawn_end,
-            'spawn_verified': self.spawn_verified,
+            'spawn_verified':
```
LegitDongo, line 163: We should change this to `verified` or change the rest to include `spawn_` in front of them for consistency

LegitDongo (changes requested):
Rest looks good ✅ 
After this is changed and we get a test then I'm cool with a merge.

LegitDongo:
Adding docs is also required, sorry I forgot in the review.

LegitDongo (approved):
✅

PR #711, opened 2020-01-23
Add is_exclusive field support for eggs

JimBim89:
## Description
This adds functionality to show if an egg has an exclusive raid or not.
Also has a filter to be able to put those eggs into different alarms.

I also added some missing dts to docs while updating it for this PR.

## Type of Change

## Motivation and Context
MAD can already send this information, so it only makes sense to use it.
I have set up an extra channel for ex eggs after nearly missing one earlier.

## How Has This Been Tested?
Tested with latest PokeAlarm and MAD.

## Wiki Update


LegitDongo (approved):
Looks good

PR #712, opened 2020-01-26
Add team emoji

x3r0x:
## Description

This adds an option to display the color of the team, which is in control of an egg or raid event. Also adds an option for the team currently and previously controling a gym.
## Type of Change

## Motivation and Context

At the moment PA doesn't have an option to display the corresponding team color for egg, raid and gym events. Some people like an emoji more than text.

## How Has This Been Tested?

Tested with latest PA and MAD.
## Wiki Update


LegitDongo:
@x3r0x do you have any qualms with changing the heart emojis to the corresponding circle ones?

Also, I'm planning on adding one more DTS on top of this PR, color names.

x3r0x:
I suspect this is now in conflict with the PVP PR.

LegitDongo (approved):
Tested it out, works great
Only thing I'm not sure about is the translations

We will probably have to restart Travis for this based on the current tests status

Great work!

kvangent (approved):

PR #716, opened 2020-04-10
Update installation.rst

TechG3n:
## Description

When python3 is used we need to use virtual env to avoid problems with other projects like MAD or RocketMAD.

## Type of Change

## Motivation and Context

Saw the new issue and remembered my problems without virtual env.

## How Has This Been Tested?

Used it with MAD

## Wiki Update

### DIFF
--- a/docs/getting-started/installation.rst
+++ b/docs/getting-started/installation.rst
@@ -24,6 +24,20 @@ This guide assumes the following:
    they change the encoding on the file and prevent it from loading correctly.
 
 
+Python Virtual Environment
+-------------------------------------
+
+A virtual environment is a way to install python packages in a different location to avoid potential version conflicts with other software like MAD or RocketMAD. It’s like a standalone version of python, independent of your “normal” python. Install it with:
+
+``apt install python-virtualenv``
+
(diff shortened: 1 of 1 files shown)


LegitDongo:
@kvangent I'm not familiar with venv, would you check this?

docs/getting-started/installation.rst
```diff
@@ -24,6 +24,20 @@ This guide assumes the following:
    they change the encoding on the file and prevent it from loading correctly.
 
 
+Python Virtual Environment
+-------------------------------------
+
+A virtual environment is a way to install python packages in a different location to avoid potential version conflicts with other software like MAD or RocketMAD. It’s like a standalone version of python, independent of your “normal” python. Install it with:
+
+``apt install python-virtualenv``
```
kvangent, line 32: this probably needs to include `sudo`? I think it might be better to avoid specifics and just point to the docs in case it changes (https://virtualenv.pypa.io/en/latest/user_guide.html)

kvangent (changes requested):
Any reason not to recommend [pyenv](https://virtualenv.pypa.io/en/latest/user_guide.html) as well?

kvangent:
Closing this, as it has gone stale.

PR #722, opened 2020-06-19
README Docs Ref master -> dev

JabLuszko:
## Type of Change

## Wiki Update


LegitDongo (approved):
Approved since we are using dev as our primary branch instead of master (at least for the time being)

PR #724, opened 2020-06-30
Add option to disable log file

Chefkeks:
## Type of Change

## Motivation and Context

Currently PA does allows logging to file and syslog. Logging to syslog can be disabled via --quiet / -q but there is no option to disable file logging. Running PA as a service and letting syslog / logrotate handle the logs, PA is still logging to the given log file. There is no option to disable that. 

## How Has This Been Tested?

Tested/Using on a production environment and works as expected. Stopped PA service, edited start_pa.py file + config/config.ini, started PA service and checked if there is still a log file. Changed the option back to another value to make sure logging is only affected if either none or None are specified. Worked like a charm to me.

## Wiki Update

- [X] The changes for config.ini.example and the read-the-docs / wiki / manual files are included too.

### DIFF
--- a/config/config.ini.example
+++ b/config/config.ini.example
@@ -17,7 +17,7 @@
 #debug                          # Enable debug logging (default='False)
 #quiet                          # Disable output to stdin/stdout.
 #log-lvl: 3                     # Verbosity of the main logger (default=3)
-#log-file: logs/pokealarm.log   # File path of the main logger (default='logs/pokealam.log')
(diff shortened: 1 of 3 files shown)


Chefkeks:
If you guys prefer another implementation, eg. as an own argument, just let me know and preferably tell me the desired arg name. Will adjust the PR then accordingly.

LegitDongo:
@kvangent what do you think?
I'm not very familar with configargparse

kvangent (approved):

PR #726, opened 2020-07-10
Fix logs for manager_name

JabLuszko:
## Description
Do not use hardcoded manager_0, just use the actual name.

## Type of Change

## Motivation and Context
Because, reasons.

## How Has This Been Tested?
Not tested, should work.

## Wiki Update


LegitDongo (approved):
LGTM

PR #732, opened 2020-08-21
Move to Python 3

LegitDongo:
## Description
Complete switch to Python 3.7+

Note: Python 2.7 will no longer be supported and anything < Python 3.7 won't work

The version is based on the packages required versions and some of the new python keywords

Also, you will probably want to set up a venv for this

Anybody who previously used Twilio needs to manually update their library:
`pip install twilio==6.45.1`

***Warning*** - copy your cache files before starting PA with this PR! There may be some unexpected edits to them

***This will require extensive testing***

## Type of Change

## Motivation and Context
Python 2 was discontinued as of January 2020.. We're a little late

## How Has This Been Tested?
I've tested on 3.7 on Windows, tested all types using the webhook tester
No prod test yet

## Wiki Update

Wiki changes will come later; we already have a PR about setting up venv; will either add to that or open a new PR for docs

Thanks Sn0opy for the help ❤️


PokeAlarm/Alarms/__init__.py
```diff
@@ -27,4 +27,4 @@ def alarm_factory(mgr, settings, max_attempts, api_key):
         from PokeAlarm.Alarms.Twitter import TwitterAlarm
         return TwitterAlarm(mgr, settings)
     else:
-        raise ValueError("%s is not a valid alarm type!".format(kind))
+        raise ValueError(f"{kind} is not a valid alarm type!")
```
kvangent, line 30: f-strings weren't introduced until python 3.6 - so you need to move your min version up

PokeAlarm/Load.py
```diff
@@ -113,10 +113,10 @@ def parse_filter_section(section):
     defaults = section.pop('defaults', {})
     default_dts = defaults.pop('custom_dts', {})
     filter_set = OrderedDict()
-    for name, settings in section.pop('filters', {}).iteritems():
-        settings = dict(defaults.items() + settings.items())
-        local_dts = dict(default_dts.items()
-                         + settings.pop('custom_dts', {}).items())
+    for name, settings in section.pop('filters', {}).items():
+        settings = dict(list(defaults.items()) + list(settings.items()))
```
kvangent: Have you verified this works right? I'm not sure of the behavior of generator -> `List<Tuple>` -> dict

kvangent: Looks like it might be okay - dict seems to be able to take a list of tuples ok

Might be more efficient/cleaner to use [`itertools.chain`](https://docs.python.org/3/library/itertools.html#itertools.chain) instead of converting though

PokeAlarm/Manager.py
```diff
@@ -523,7 +523,7 @@ def setup_in_process(self):
             logging.getLogger().setLevel(logging.DEBUG)
 
         # Conect the alarms and send the start up message
-        for alarm in self._alarms.values():
+        for alarm in list(self._alarms.values()):
```
kvangent: This looks wrong - `dict.values()` should already return a list

PokeAlarm/Manager.py
```diff
@@ -746,10 +749,10 @@ def process_stop(self, stop):
         rules = self.__stop_rules
         if len(rules) == 0:  # If no rules, default to all
             rules = {"default": Rule(
-                self._stop_filters.keys(), self._alarms.keys())}
+                list(self._stop_filters.keys()), list(self._alarms.keys()))}
```
kvangent: same with `dict.keys` - should already be a list

start_pokealarm.py
```diff
@@ -23,14 +22,12 @@
...
-from PokeAlarm.Utils import get_path, parse_unicode, parse_boolean
+from PokeAlarm.Utils import get_path, parse_boolean
 from PokeAlarm.Load import parse_rules_file, parse_filters_file, \
     parse_alarms_file
 
 # Reinforce UTF-8 as default
-reload(sys)
-sys.setdefaultencoding('UTF8')
-
+os.environ["PYTHONIOENCODING"] = 'utf-8'
```
kvangent: You can remove this line altogether

tests/filters/test_monster_filter.py
```diff
@@ -1,16 +1,15 @@
...
+import os
 import unittest
-import sys
 import PokeAlarm.Filters as Filters
 import PokeAlarm.Events as Events
 
 # ToDo: Find a better way
 # Reinforce UTF-8 as default
-from tests.filters import MockManager, generic_filter_test, full_filter_test
+os.environ["PYTHONIOENCODING"] = 'utf-8'
```
kvangent: drop this as well

tools/convert_alarms_file.py
```diff
@@ -28,48 +28,48 @@ def exchange_set(settings):
...
+    except ValueError:
+        print("Encounter a ValueError while loading the file. Make sure "
+              "your file is in the correct JSON format then try again.")
+    except IOError:
+        print("Unable to find file. Please check that the file exists "
+              "and the program has the proper privileges then try again.")
     except Exception as e:
-        print ("An unexpected error as occurred: {} - {}"
-               "".format(type(e).__name__, e))
+        print(("An unexpected error as occurred: {} - {}"
```
kvangent: looks like an extra layer of parenthesis here

tools/pvp.py
```diff
@@ -12,15 +12,15 @@ def __init__(self, pa_root):
 
         multipliers = utils.get_cp_multipliers()
 
-        for json_mon in stats.keys():
+        for json_mon in list(stats.keys()):
```
kvangent: Shouldn't need list

tools/pvp.py
```diff
@@ -12,15 +12,15 @@ def __init__(self, pa_root):
...
             for limit in [1500, 2500]:
                 highest, lowest = self.spreads(
                     limit, json_mon, min_level(limit, json_mon),
                     max_level(limit, json_mon), multipliers, stats)
                 stats[json_mon]["{}_product".format(limit)] \
                     = highest["product"]
-                print("{}: highest product at {}: {}".format(
-                    json_mon, limit, highest['product']))
+                print(("{}: highest product at {}: {}".format(
```
kvangent: extra parenthesis here

tools/webhook_test.py
```diff
@@ -8,12 +8,15 @@
 import pickle
 from glob import glob
 
+# Reinforce UTF-8 as default
+os.environ["PYTHONIOENCODING"] = 'utf-8'
```
kvangent: drop this

kvangent (changes requested):

crhbetz:
No issues so far, running 24hrs on production (2 PA instances / 34 MAD devices total). I'll stick with it and report issues asap in case I see anything.

kvangent (approved):

PR #734, opened 2020-09-06
Switch CI to Github Actions

kvangent:
Github actions is likely more reliable than TravisCI these days.


LegitDongo (approved):
I'm not super familiar with github CI, but based off what I think is happening this should be fine
My only question is how we want to add tests/support with #732 
I'd think we'd only add 2.7 with this PR, then with #732 we'd remove 2.7 and add 3.7 & 3.8

Your call

kvangent:
@LegitDongo My plan was to just have you disable 2.7 in #732 - everything's already failing because of some lint failures

PR #740, opened 2020-10-18
Python 3 Pushbullet Alarms Fix

LegitDongo:
## Description
Replaces PushBullet API (which doesn't support python 3.7) with flat requests

## Type of Change

## Motivation and Context
Makes PushBullet alarms work for the newest python 3 move

## How Has This Been Tested?
Locally, windows 10

## Wiki Update


PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
```diff
@@ -182,19 +184,59 @@ def invasion_alert(self, invasion_info):
...
+        if res.status_code == requests.codes.ok:
+            self._log.debug('Notification successful '
+                            f'(returned {res.status_code})')
+        else:
+            self._log.debug(f'PushBullet response was {res.content}')
+            raise requests.exceptions.RequestException(
+                f'Response received {res.status_code}, webhook not accepted.')
+
+    def update_channels(self):
+        self.__channels = {}
```
kvangent: avoid setting this until the end - if some does go wrong, you don't want to break the existing list of channels.

PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
```diff
@@ -182,19 +184,59 @@ def invasion_alert(self, invasion_info):
...
+        response = requests.get(
```
kvangent: We should use a Session here: https://requests.readthedocs.io/en/master/user/advanced/#session-objects

Initialize in `__init__` and you can set the access-token header by default in a single place

PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
```diff
@@ -182,19 +184,59 @@ def invasion_alert(self, invasion_info):
...
+            'https://api.pushbullet.com/v2/channels',
+            headers={
+                'Access-Token': self.__api_key
+            }, timeout=30)
+        if response.ok is True:
+            self.__channels = response.json()['channels']
```
kvangent: What happens if `channels` isn't present?

PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
```diff
@@ -182,19 +184,59 @@ def invasion_alert(self, invasion_info):
...
+            self._log.debug('Detected the following PushBullet channels: {}'
+                            .format(self.__channels))
+        else:
```
kvangent: It's better to avoid branching the decision tree here. If you do:
```python
if not response.ok:
    // handle error

//handle normally
```

This is easier for people to read because it reinforces the non-error path as "default" - and error handling is more of a "if something goes wrong, handle it".

kvangent (changes requested):

PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
LegitDongo: This was a holdover from my basis off of the SlackAlarm

Removing this since it isn't needed from what I can tell

kvangent (approved):

PR #748, opened 2020-11-23
Update pvp.py Tool

Audidriver13:
## Description

The calculation of the pvp products is wrong in the tool. The stamina can only take an int-value.

## Type of Change

## How Has This Been Tested?
Manual calculation was used for testing

## Wiki Update


crhbetz:
As also explained in #749 , this is correct and should be merged 😉

LegitDongo (approved):
Someday I will make this code prettier, but I'll try to verify better when I do 😄

PR #749, opened 2020-11-23
Fix stat product calculation

crhbetz:
## Description
It has been brought to my attention that using the pvp.py tool yields wrong stat product numbers. Upon inspection, I found this was because the rounding of the stamina value had been removed during a "code cleanup" 😉 

This PR fixes that issue.

## Type of Change

## Motivation and Context
Simple bug fix.

## How Has This Been Tested?
Verified the calculations to match trusted sources again (http://pogostat.com/, poraclePvpHelper plugin 😉 )

## Wiki Update


crhbetz:
@Audidriver13 played me good here. 😆

PR #751, opened 2020-12-04
Changes pvp data

Devreugkx:
## Description
Changed Generation 6 base stats as they were outdated.
Changed cp_multipliers to more accurate versions, ones that are also used by major pvp_websites
Changed max stat products for GL and UL accordingly.

## Type of Change

## Motivation and Context
Some data was outdated or could have been more updated/accurate

## How Has This Been Tested?
Hand picked and tested some Pokémon

## Wiki Update


boby360:
Has this been tested? the # symbols dont break anything?

Devreugkx:
As

PR #759, opened 2020-12-20
Updated CPM's from Pokebattler and recalculated Base_Stats 

Audidriver13:
## Description

Pokébattler updated today their cpm values. Gostadium will follow.
Turns out the cpms were incorrect at the 8th decimal place. 
The base_stats were recalculated. The changes are minimal and only important in marginal cases.

## Type of Change

## Wiki Update


LegitDongo (approved):
LGTM

PR #760, opened 2020-12-22
Cache geocoding results in manager object cache

d12fk:
## Description
The move of the cache makes restarts without losing
cached geocoding results possible; if a disk cache is used.
Previously the cache was always in-memory only.

Cached results expire after one month now.

The new 'gmaps-cache-fuzz' option can be used in the config
to flatten re-lookup spikes that could happen if a lot of
requests were made at the same day, i.e. during an event.
It randomly increases cache expiration to up to the configured
value in days (defaults to 0).

## Type of Change

## Motivation and Context
Makes it possible to restart without losing gmaps results.

## How Has This Been Tested?
The patch is running in production for two weeks and shows the
expected results with several restarts in the meantime.

## Wiki Update
The wiki seems to be dead, so either remove this section or update the URL.

### DIFF
--- a/PokeAlarm/Cache/Cache.py
+++ b/PokeAlarm/Cache/Cache.py
@@ -35,6 +35,8 @@ class Cache(object):
         self._quest_reward = {}
         self._quest_task = {}
         self._quest_last_modified = {}
+        self._loc_geocode = {}
+        self._loc_rev_geocode = {}
 
     def monster_expiration(self, mon_id, expiration=None):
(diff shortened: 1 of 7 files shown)


kvangent:
Hi @d12fk,

Thanks for taking the time to open a PR. Unfortunately, I have to decline this feature, as it's in violation of the Google Maps TOS, which doesn't allow long-term caching of results. Using this PR may result in legal action against this project or it's users.

PR #762, opened 2021-01-23
Add Evolution DTS

TiMXL73:
## Description
Add Evolution DTS for RaidEvent

## Type of Change

## Motivation and Context
New dts added so that alarms can display the evolution for raids if applicable. Those terms are language specific so the information is stored locales's json.
```
<evolution>
<evolution_or_empty>
<evolution_id>
<evolution_id_2>
<evolution_id_3>
```

## How Has This Been Tested?
It's been running locally for a while without issues (latest dev / python3)

## Wiki Update
RST is updated in this pull request.


LegitDongo (approved):
LGTM
I had a version that had some nicer information that I was planning to add but my time is limited unfortunately :(

Thanks for the PR ❤️

PR #766, opened 2021-03-28
Fix DTS for <great_url> and <ultra_url>

mikosoby:
## Description

The current states of `<great_url>` and `<ultra_url>` produces a link to stadiumgaming.gg with incorrect value for "Minimal IV" and a non existing for "Use of Best Buddy". This is likely the old usage from the preLvl 50 times.

This fix now uses the correct parameters for "Minimal IV" and "Level Cap".

Old URL:
```
https://www.stadiumgaming.gg/rank-checker?pokemon=Medicham&league=1500&att_iv=15&def_iv=15&hp_iv=15&min_iv=0&include-best-buddy=false
```
OLD

New URL:
```
https://www.stadiumgaming.gg/rank-checker?pokemon=Medicham&league=1500&att_iv=15&def_iv=15&hp_iv=15&min-iv=0&levelCap=50
```
NEW

## Type of Change

## Motivation and Context

The current state produces an URL that selects Level 40 as the max level for Pokémon. This fixes that to Level 50

## How Has This Been Tested?

I've just update my local copy of PA and waited for an event that produced an output to Discord and clicked on the link there.

## Wiki Update

### DIFF
--- a/PokeAlarm/Events/MonEvent.py
+++ b/PokeAlarm/Events/MonEvent.py
@@ -262,8 +262,8 @@ class MonEvent(BaseEvent):
                     'att_iv': self.atk_iv,
                     'def_iv': self.def_iv,
                     'hp_iv': self.sta_iv,
-                    'min_iv': '0',
-                    'include-best-buddy': 'false'
+                    'min-iv': '0',
+                    'levelCap': '50'
                 }),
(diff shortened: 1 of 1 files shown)


LegitDongo (approved):
LGTM

PR #768, opened 2021-04-09
add venue support for telegram alarm

AstAst42:
## Description
Adds possibility to use venues in telegram alarms. Venues are very similar to maps, but they can have additional text just below the text. This makes it easier to understand which "map" (venue) message belongs to which mon text message and makes it easier to share the alarm by simply sharing the venue only. You can edit the venue text like all other texts but keep in mind that the space is limited.

image

This text is also shown after pressing on the venue if you want to use navigation.

image

## Type of Change

## Motivation and Context

This makes it easier to find the Mons you are looking for because you can combine mon values with a map and improves sharing.

## How Has This Been Tested?

I used this as a local change on a server for 3 small cities, had it run for months and had no problems.

## Wiki Update

### DIFF
--- a/PokeAlarm/Alarms/Telegram/TelegramAlarm.py
+++ b/PokeAlarm/Alarms/Telegram/TelegramAlarm.py
@@ -24,7 +24,8 @@ class TelegramAlarm(Alarm):
 
     Alert = namedtuple(
         "Alert", ['bot_token', 'chat_id', 'sticker', 'sticker_url',
-                  'sticker_notify', 'message', 'message_notify', 'venue',
+                  'sticker_notify', 'message', 'venue_message',
+                  'message_notify', 'venue',
                   'venue_notify', 'map', 'map_notify', 'max_attempts',
                   'web_preview'])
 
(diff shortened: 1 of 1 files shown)


LegitDongo:
This seems like it would cause unexpected results for those who use current message functionality with the venue option on

Can someone confirm that this works as expected even without the venue message option?

PokeAlarm/Alarms/Telegram/TelegramAlarm.py
```diff
@@ -24,54 +24,63 @@ class TelegramAlarm(Alarm):
...
+                  'sticker_notify', 'message', 'venue_message',
+                  'message_notify', 'venue',
                   'venue_notify', 'map', 'map_notify', 'max_attempts',
                   'web_preview'])
 
     _defaults = {  # No touchy!!! Edit alarms.json!
         'monsters': {
             'message': "*A wild <mon_name> has appeared!*\n"
                        "Available until <24h_time> (<time_left>).",
+            'venue_message': "",
```
billyjbryant, line 36: Wouldn't the proper thing here be to set `venue_message` to `null` rather than an empty string?

billyjbryant:
Is it possible to disable this in settings so that Venues show only if they are either enabled or not disabled?

billyjbryant (changes requested):
Please address the comments on this PR

billyjbryant:
This PR appears to be abandoned, please re-open and address comments if you wish to introduce this feature.

PR #772, opened 2021-08-13
Gen 8 Update for the upcoming event

Audidriver13:
## Description

Part 3 of the hyperbonus event starts on August 20th with the new Generation 8 Pokémon. In preparation for this, all values of Gen8 were added, the PVP stats were recalculated (base_stats.json) and the Pokémon names were added in the locales (de, en, fr). The (unnecessary?) Information on the height and weight of the Pokémon was omitted due to time constraints.

Note: The calculation for generation 7 is still incorrect because no correct base values are available yet.

## Type of Change

## Wiki Update


Audidriver13:
If this PR is accepted, PR # 771 can be closed because it has been implemented here.

TiMXL73:
Gen 8 is going to be live in less than 6h in New Zealand and PA still doesn't support Gen 8 until this or an updated variation of it is merged.

NOTE: The updated locale json are using TABS instead of SPACES for their indentation. This should eventually be fixed along with adding dummy/temporary values for the other locale languages to avoid errors (ie, use english for now and fix later?).

Audidriver13:
@TiMXL73 have implemented your notes :) thanks for the hint

locales/en.json
```diff
@@ -4590,7 +4590,7 @@
     "769": "Sandygast",
     "770": "Palossand",
     "771": "Pyukumuku",
-    "772": "Type: Null",
+    "772": "TypeNull",
```
LegitDongo: This, along with the Tapu mons, doesn't look correct, is this how it is in-game?

LegitDongo (changes requested):
Just one complaint

Thanks for the amendments, I had planned on making them myself but have found my time lacking of late

locales/en.json
Audidriver13: The correct name is "Type: Null" (https://www.pokemon.com/us/pokedex/type-null), but I changed it to "TypeNull" to be on the safe side so that there are no errors.

Audidriver13: Oh. The Tapu Mons is a typical case of "finding and replacing spaces". I'll put the space right back in

LegitDongo: Special characters and spaces aren't a problem for PA, the main problem there lies when the encoding is outside of UTF-8, because of how the json files themselves are encoded
This is supplemented elsewhere by giving python unicode strings (the "\u..." you see here and there) to represent the properly encoded string representation

Back on topic, references to this are usually done in a string, such as in filters: `monsters: ["Type: Null"]`, so spaces aren't a problem

Audidriver13: Tapu mons changed with spaces. The Type: Null mon ... you can decide, should I use the original name (Type: Null) or leave it that way (TypeNull)?

LegitDongo: change it to `Type: Null` please 😄

Audidriver13: done :)

LegitDongo (approved):
This will do for now, will go back over this some other time

PR #774, opened 2021-08-22
Update Alolan Raticate UL product+ Formatting

Audidriver13:
## Description

Update of the 2500_product of alolan raticate, although this change has no effect because Alolan Raticate cannot reach 2500 CP (max cp: 1927). Just for correctness

## Type of Change

## Wiki Update


LegitDongo (approved):
Looks like it's been used a few times now, so LGTM

PR #784, opened 2021-10-23
Update Stats for Pumpkaboo

Audidriver13:
## Description

Update stats for Pumpkaboo and Gourgeist.
Since no forms are possible in PA for the PVP calculation, I took the Super form, as this is the best in PVP.

## Type of Change

## Wiki Update


LegitDongo (approved):
Haven't checked stats, will take your word for it

PR #786, opened 2022-01-25
Add monster types filter

Pontiky:
## Type of Change

## Motivation and Context

Adding a monster types filter

## How Has This Been Tested?

Tested locally with a new `filters.json` below and everything worked fine.

```json
{
  "monsters":{
    "enabled": true,
    "defaults":{
    },
    "filters":{
      "filter_dark_pokemons":{
        "types": ["Ténèbres"]
      },
      "filter_dragon_normal_flying_pokemons":{
        "types": ["Dragon", "Normal", "Vol"]
      }
    }
}
```

## Wiki Update

Need to add a new line in Filters -> Monsters -> General table named with fields:
`Parameter: types`
`Description: Array of allowed monster types`
`Example: ["Dragon", "Dark"]`


billyjbryant:
Changes to locale translations should be in a different PR than changes to functionality.

PokeAlarm/Filters/BaseFilter.py
```diff
@@ -180,7 +180,10 @@ def __init__(self, limit, eval_func, attr_name):
 
     def __call__(self, filtr, event):
         value = getattr(event, self._attr_name)  # event.event_attr
-        if Unknown.is_(value):
+        if (type(value) == list):
+            if Unknown.is_(*value):
```
billyjbryant, line 184: Explain what is going on with this new section of code. What are these nested if statements doing or not doing? Has this been tested, if so, how, can you show the test results and what was required to produce them?

billyjbryant (changes requested):
You should write tests to insure your filters do function as intended. Tests for Monster Filters are included here: https://github.com/pokealarm/pokealarm/blob/dev/tests/filters/test_monster_filter.py

billyjbryant:
PR is also failing Flake8

PokeAlarm/Filters/MonFilter.py
```diff
@@ -215,6 +221,10 @@ def to_dict(self):
         if self.max_dist is not None:
             settings['max_dist'] = self.max_dist
 
+        # Pokemon types
+        if self.type_ids is not None:
+            settings['type_ids'] = self.type_ids
```
billyjbryant, line 227: Please include a test for this filter in https://github.com/pokealarm/pokealarm/blob/dev/tests/filters/test_monster_filter.py to ensure it properly works and evaluates before merger.

PokeAlarm/Filters/BaseFilter.py
Pontiky: Since now, all the evaluated values were a single integer and this one was given to `Unknown.is_()`.
The new value `types` is a list of 1 or 2 int corresponding to the types_ids of the monster.
The new change passes the list as multiple args to `Unknown.is_()` using the `*list` syntax.

Pontiky:
> You should write tests to insure your filters do function as intended. Tests for Monster Filters are included here: https://github.com/pokealarm/pokealarm/blob/dev/tests/filters/test_monster_filter.py

Done and committed the test
```
Ran 30 tests in 0.254s

OK
```

billyjbryant (approved):

PR #787, opened 2022-01-30
Hisui update

Audidriver13:
## Description

Update of new Hisui Voltorb form

## Type of Change

## Wiki Update


billyjbryant:
Code change looks good but https://monsterimages.tk/regular/monsters/100_2728.png does not resolve and instead reverts to 100_000.png. Looks like the base images haven't been updated yet.

billyjbryant (approved):

billyjbryant (changes requested):
This change needs to be introduced into all Locales at one time. Please include translations for `ko`, `pt` and `zh_hk` as wel.

Audidriver13:
> This change needs to be introduced into all Locales at one time. Please include translations for `ko`, `pt` and `zh_hk` as wel.

Updated the remaining languages, but not in local language (ko and zh_hk). But it was often used that way.

billyjbryant (approved):

PR #778, opened 2021-09-18
Display monsters (ditto)

LegitDongo:
## Description
Add new MAD-based "display" monster information
tl;dr, the monster that gets shown in the overworld when it's a ditto

Note: these keys are subject to change since nothing has really been approved

It adds some nice DTS

`display_mon_id` - the ID of the monster being displayed
`display_mon_name`  - the name of the displayed monster (based on locale)
`display_mon_id_3` - a pre-padded zero 3-digit number representing the displayed monster ID
`display_mon_id_2` - a pre-padded zero 2-digit number representing the displayed monster ID
`display_costume_id` - the ID of the costume the displayed monster has
`display_costume_id_2` - a pre-padded zero 2-digit number representing the displayed monster's costume ID
`display_costume_id_3` - a pre-padded zero 3-digit number representing the displayed monster's costume ID
`display_costume` - the full name of the displayed monster's costume (based on locale)
`display_form_id` - the ID of the displayed monster's form
`display_form` - the name of the displayed monster's form (based on locale)
`display_form_id_3` - a pre-padded zero 3-digit number representing the displayed monster's form ID
`display_form_id_2` - a pre-padded zero 2-digit number representing the displayed monster's form ID
`display_gender` - the greek symbol representing the displayed monster's gender

Since we need a way to use the "display" monster image for the default image provider, I added some alarm keys for monsters too

Discord: `use_display_icon` and `use_display_avatar`
FB: `use_display_image`
Slack: `use_display_icon`
Telegram: `use_display_sticker`

!! this gets overridden if the key exists for a custom image provider !!

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (would cause existing functionality to change)

## Motivation and Context
Some people want to know which monster to click on

## How Has This Been Tested?
win 10, python 3.8, discord only

## Wiki Update
- [x] This change requires an update to the Wiki.
- [ ] This change does not require an update to the Wiki.

Wiki changes to come later based on team discussion, see above for full listing of new keys

PR #791, opened 2022-02-04
Revert "Display monsters (ditto) (#778)"

billyjbryant:
## Description
This PR reverts #778 which introduced an un-intentional bug that breaks Telegram alerts. 

## Type of Change

## Motivation and Context

Telegram was broken by #778, @LegitDongo is working on a fix. 

## Wiki Update


LegitDongo (approved):
fine with me, will get a fix + new PR soon

PR #794, opened 2022-02-05
Add candy and stardust costs for PvP

### LINKED ISSUE #728: Calculate Candy and Stardust for PvP-Mon
## Description
It would be nice to have the option to display/calculate the needed candy and stardust for the evolution of the mon.

Candy and stardust are fixed per level therefore it should be possible to calculate it.
See https://rankedboost.com/pokemon-go/stardust/ as reference.

Pontiky:
## Description

Added new DTS `<great_stardust>`, `<great_candy>`, `<ultra_stardust>`, `<ultra_candy>` to get the cost to power up a Pokemon for PvP to its final level/evolution/CP.
The candy cost is taking evolution and XL candies into account.

## Type of Change

## Motivation and Context

This change was requested and will close #728.
The new DTS can be used by players to know if it's worth to capture and power up a Pokemon for PvP.

## How Has This Been Tested?

I did multiple tests on a few Pokemons using PokeGenie app to compare values.
#### Test 1:
image

#### Test 2:
image

#### Test 3:
image

## Wiki Update

Included in this PR.


Gerbrum:
Working fine for me so far, just one small mistake I found:
`400: [130, 320, 333, 808, 714]` should be changed to `400: [129, 320, 333, 808, 714]` (Gyarados --> Magikarp)
I didn´t check all the evolution ID´s, saw the mistake as Magikarp is spawning right now.
Thanks for the great PR!

Pontiky:
> Working fine for me so far, just one small mistake I found: `400: [130, 320, 333, 808, 714]` should be changed to `400: [129, 320, 333, 808, 714]` (Gyarados --> Magikarp) I didn´t check all the evolution ID´s, saw the mistake as Magikarp is spawning right now. Thanks for the great PR!

You're right! I replaced Gyrados by Magikarp.

I also updated the evolution cost list with missing Pokemons.

PokeAlarm/Utilities/PvpUtils.py
```diff
@@ -59,6 +59,101 @@ def min_level(limit, monster):
...
+        100: [2, 5, 8, 33, 30, 42, 44, 61, 64, 67, 70, 75, 82, 93, 108, 112,
+              114, 117, 122, 125, 126, 148, 153, 156, 159, 176, 180, 188, 190,
+              193, 198, 200, 207, 215, 221, 233, 247, 253, 256, 259, 271, 274,
+              281, 288, 305, 315, 329, 349, 356, 364, 372, 375, 388, 391, 394,
+              397, 404, 430, 444, 478, 496, 499, 502, 507, 536, 541, 544, 575,
+              578, 583, 600, 603, 608, 611, 634, 651, 654, 657, 662, 670, 680,
+              723, 726, 729, 732, 737, 762, 783, 790, 811, 814, 817, 822, 825,
+              838, 857, 860, 886],
+        200: [525, 533, 588, 616, 710],
+        400: [129, 320, 333, 636, 714, 808]
```
LegitDongo: This seems like data that should go in `data/base_stats.json`, since these are specific to the monster

Either that or in it's own data file; I just don't like having readily-changing data sitting plain in an app

What do you think @billyjbryant ?

PokeAlarm/Utilities/PvpUtils.py
```diff
@@ -59,6 +59,101 @@ def min_level(limit, monster):
...
+    else:
+        return f'{candy_cost:,}'.replace(',', ' ')
+
+
+def calculate_stardust_cost(start_level, target_level):
+    start_level = float(start_level)
+    target_level = float(target_level)
+    stardust_table = [200, 400, 600, 800, 1000, 1300, 1600, 1900, 2200,
+                      2500, 3000, 3500, 4000, 4500, 5000, 6000, 7000,
+                      8000, 9000, 10000, 11000, 12000, 13000, 14000, 15000]
```
LegitDongo: Is this information subject to change in the future?

PokeAlarm/Utilities/PvpUtils.py
Pontiky: I thought about the same thing when I realized that there is a data folder with json files few hours ago. Maybe it could be a good idea.

PokeAlarm/Utilities/PvpUtils.py
Pontiky: Since the game release, it looks like they never changed the stardust and candy costs for powering up a Pokemon.
[Link to a reddit post, on July 2016](https://www.reddit.com/r/TheSilphRoad/comments/4sa4p5/stardust_costs_increase_every_4_power_ups/)

PokeAlarm/Utilities/PvpUtils.py
billyjbryant: I agree, either merged in to base-stats or its own data file in to data folder.

Pontiky: I inserted the candy cost of evolutions in `base_stats.json` and modified the evo cost function consequently.

PokeAlarm/Utilities/PvpUtils.py
billyjbryant: I would prefer to see data (even potentially long lived data) in a data file rather than hard coded into the programming.

billyjbryant (changes requested):
Please move data values to a data file and out of the programming of the application.

billyjbryant (approved):

PR #795, opened 2022-02-06
Add time filters

### LINKED ISSUE #690: Add time option to tigger alarm only of certain hours
## Description

Add time option that allows trigger alarm only of certain hours.

It's can turn off some notification at night or raid hours/days to prevent notification flood. Can also turn on some rules only on community day time.

Add option of time ranges. When time is in range, then tigger alarm.

Pontiky:
## Description

The new filters DTS (`min_time` and `max_time`) allow users to specify a time range in which an alarm can be triggered.
These DTS use the local time of the PA server.

## Type of Change

## Motivation and Context

There was no way to filter an alarm using a time range and the only workaround was to stop or reload PA using CRON.
These filters can also be used to only trigger alarms during the daytime, a community day or a spotlight hour.
This one also closes #690.

## How Has This Been Tested?

Given the current time changes depending on... the current time, I can't really write tests for this one as it would depends of the time where it's tested.

What I tried is multiples filters with different `min_time` and `max_time` values. From what I tested, I didn't find specific values where the time filters failed.

## Wiki Update

Included in this PR.

### DIFF
--- a/PokeAlarm/Filters/BaseFilter.py
+++ b/PokeAlarm/Filters/BaseFilter.py
@@ -1,5 +1,6 @@
 # Standard Library Imports
 import logging
+from datetime import datetime, timedelta
 # 3rd Party Imports
 # Local Imports
 from PokeAlarm import Unknown
@@ -73,6 +74,18 @@ class BaseFilter(object):
         # Add check function to our list
         self._check_list.append(check)
         return limit
+
+    def evaluate_time(self, min_time, max_time):
+        if min_time is None:
+            min_time = 0.0
+        if max_time is None:
+            max_time = 60.0 * 60.0 * 24.0
+
+        # Create a function to compare the current time to time range
+        check = CheckTime(min_time, max_time)
+
+        # Add check function to our list
+        self._check_list.append(check)
 
     @staticmethod
     def parse_as_type(kind, param_name, data):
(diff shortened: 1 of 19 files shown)


PokeAlarm/Filters/BaseFilter.py
```diff
@@ -191,3 +227,30 @@ def __call__(self, filtr, event):
...
+
+    def __init__(self, min_time, max_time):
+        self._min_time = min_time
+        self._max_time = max_time
+
+    def __call__(self, filtr, event):
+        now = datetime.now()
+        current_time = (now - now.replace(hour=0, minute=0,
+                        second=0)).total_seconds()
+        if self._min_time <= self._max_time:
```
billyjbryant, line 250: It feels like having a filter set with `min_time == max_time` would cause a failure in the logic dictated here. You should probably split these out into separate evaluations. 

What is the expected outcome if someone sets both filters to be the same time?

Pontiky: Logically, if you set any time to X:XX, we assumes that means X:XX:00.
Now, if you set a time range with the same minimum and maximum (which means a range from X:XX:00 to X:XX:00), the only alarm which passes has to be on the exact same second (so X:XX:00). For any other time, the filter will reject the event. So I would say the expected outcome is logic.

What do you think about this behaviour?

billyjbryant: That makes sense.

billyjbryant (changes requested):
Please write unit tests for the new filter options. You can see other prior filter tests here: https://github.com/pokealarm/pokealarm/tree/dev/tests/filters

PokeAlarm/Filters/BaseFilter.py
```diff
@@ -74,6 +75,18 @@ def evaluate_attribute(self, limit, eval_func, event_attribute):
         self._check_list.append(check)
         return limit
 
+    def evaluate_time(self, min_time, max_time):
```
billyjbryant, line 78: For unit-testing, you can optionally add a 3rd arg here `override_time` that defaults to `Null` then pass this argument as a 3rd arg to `CheckTime()` where you'd do the evaluation with the value of `override_time` being the time used instead of `datetime.now()`

Pontiky: Tweaked something to override `current_time` for unit tests. Tests passed:
```
Ran 31 tests in 0.260s
OK
```

billyjbryant (approved):
Looks good, merging

PR #800, opened 2022-02-12
Add new moves portuguese translation

lzibeti:
## Description

Add new moves portuguese translation and add move "lock-on".
## Type of Change

## Motivation and Context

The "lock-on" move always appeared as unknown and looking at the files I saw that it was missing. I took the moment to update the file in Portuguese, updating with the new moves.

## How Has This Been Tested?

Translation is being used on my server, in production environment, sending information to 2 Telegram groups without error.
## Wiki Update


locales/pt.json
```diff
@@ -3805,7 +3805,77 @@
     "278": "Rosnado",
     "279": "Mastigada",
     "280": "Jogo Sujo",
-    "281": "Poder Oculto"
+    "281": "Poder Oculto",
```
billyjbryant, line 3808: Please fix spacing for all entries below this one to ensure uniformity of the file.

billyjbryant (changes requested):

locales/pt.json
lzibeti: Sorry, I didn't know it generated this difference using notepad++. I'll fix it.

lzibeti: Done

billyjbryant (approved):

Issue #770, opened 2021-06-27
Fetch Base Stats and Move Info


LegitDongo:
## Description
Get the base stats and move info based on an interval or on request
Allow the fetch URL to be changed

## Why is this enhancement necessary or useful?
This changes enough that we don't want to make a git commit every time they change
Also, we are pretty bad with keeping up with it

## How can this enhancement be accomplished?
A few config options, a http request, and change some data around

billyjbryant:
@LegitDongo is allowing the URL to be configurable a hardfast requirement of this FR? Seems that changing the URL to something different that might have a different file format or data structure would fail unless you also had a way of specifying the schema of the file, which would be difficult. If this is something we can live with not being configurable by the end user, I believe we can call this request closed as #802 is already merged which pulls the data dynamically.

PR #802, opened 2022-02-12
Fetch mon data automatically and form support

### LINKED ISSUE #683: Forms show wrong type / weather / cp values
## Troubleshooting Checklist

- [x] My Scanner install is at the most recent version.
- [x] I've already checked the [FAQ](https://github.com/RocketMap/PokeAlarm/wiki/faq) and the [Wiki](https://github.com/RocketMap/PokeAlarm/wiki) already.

## Problem Description

I'm glad PA started to support form ids. The form name and pictures do properly show, but types (also affecting weather) and cp values are currently wrongly displayed/calculated.

Pontiky:
## Description

This is a big pull request whose the aim is to fetch the Pokemon data automatically when PokeAlarm  is starting. This includes a lot of changes from how the data is read by the PA modules.

### Update behaviour:
- At startup of PA, the new function `check_for_update()` checks the presence of `data/.data_version` by calling the github API about the SHA of the last commit of `master-latest-everything.json` (from the repo where PoracleJS is updating its data). If the local and remote signatures don't match, the local Pokemon data is fetched to get a local copy in `data/pokemon_data.json`
- These two files were added in `.gitignore`.
- With the above and below changes, `data/base_stats.json` has been removed.

Masterfile used to fetch data: https://github.com/WatWowMap/Masterfile-Generator/blob/master/master-latest-everything.json
Saved as `data/pokemon_data.json`.
This is from the same repo as PoracleJS but not exactly the same file because we need a lot of different informations and the above file was the most complete.

### Changes from `Utils` module:
- All the functions which gather Pokemon data have been updated to match the new data format of `data/pokemon_data.json`. This includes minors changes for the following informations: weight, height and stats.
- The function `get_evolutions()` has been rewritten to match the new format and to support the Pokemon forms particularities (e.g. `Zigzagoon -> Linoone` or `Galarian Zigzagoon -> G. Linoone -> G. Obstagoon`). It now supports a 2nd parameter named `form_id`.
- The function `get_base_types()` has been rewritten to match the new format and to support the Pokemon forms particularities (e.g. `Rattata (Normal type)` or `Alolan Rattata (Normal/Dark types)`). These changes have been applied to `is_weather_boosted()` function.
- The function `get_base_stats()` has been rewritten to match the new format and to support the Pokemon forms particularities (e.g. `Raticate (atk: 161, def: 139, sta: 146)` and `Alolan Raticate (atk: 135, def: 154, sta: 181)`)
- Stats products are now saved in a different data file (`data/stat_products.json`) because that was the only stats used in the old `data/base_stats.json`.
- A new function `get_raw_form_names()` has been added to retrieve all the forms automatically from the last updated `data/pokemon_data.json`. There is no more need to commit new forms as they are processed automatically from this new file.

### Changes from `Locale` module:
- The way to how form names are translated has been changed in locale files. See the new `form_names` key. The form names from `data/pokemon_data.json` (obviously in English) are used by default but can be modified by the `form_names` key (even in `locales/en.json`).
- The Fairy and Dark ID types have been swapped to match `data/pokemon_data.json`.

### Changes from `Events` module:
- Some minor changes for `MonEvent`, `QuestEvent` and `RaidEvent` to give the form id to the functions which need it (pokemon types, PvP stats, weather boost)

### Changes from `PvpUtils` module:
- Some generic functions have been moved to `Utils` module like `calculate_cp()`, `get_cp_multipliers()` and a few other which are not directly linked to PvP compared to the remaining functions `pokemon_rating()` and `get_pvp_info()`. These changes have have been pushed here because `Utils` module got a really big update.

## Type of Change

This PR could change the behaviour of current filters as `form_id` could update types or evolutions.
Some form names have also been changed following the names used in `data/pokemon_data.json`. They are not used in current filters as we have to use form ID instead of names. These changes should not impact current configured filters.

## Motivation and Context

These changes are highly required because they bring automatic fetching of new data without any commit needed (except translations). The new `data/pokemon_data.json` is also bring the support of form particularities which is an other major improvement for PA.

This request will update #770 (high priority). Only fetching `move_data.json` has to be implemented after this PR.
This one will also partially address #683 by starting to support form particularities.

## How Has This Been Tested?

I tested these changes with a live working PA for a couple of days and I didn't notice any new error or crash. I also made a lot of debug and tests with `tools/webhook_test.py` to ensure that form particularities such Pokemon evolutions or types are correctly supported.

## Wiki Update


billyjbryant (changes requested):
You need to squash your commits, you are pulling in a lot of commits here that are already committed to the repo. This can sometimes cause issues with merging. Typically this happens when you do an improper merge & rebase.

CarlosVMX1:
I had been using this PR for almost 5 days and it works correct.

billyjbryant (changes requested):
Please move all locale changes into a separate PR

tools/pvp.py
```diff
@@ -2,32 +2,54 @@
...
+import requests
 
 
 class PVP:
     def __init__(self, pa_root):
-        with open(pa_root + "/data/base_stats.json", "r") as json_data:
-            stats = json.load(json_data)
-            json_data.close()
+        # Fetch pokemon data
+        master_file = "https://raw.githubusercontent.com/WatWowMap/" \
```
billyjbryant, line 11: We should never fully rely on loading a remote file. There should be a way of having a backup, such as retrieving this file and caching it, then referring to the cached version if the remote file becomes unavailable. Another option is to add this Repo as a sub-repository within the PA repo that is pulled anytime you perform a `git pull`.

tools/pvp.py
```diff
@@ -2,32 +2,54 @@
...
+            "Masterfile-Generator/master/master-latest-everything.json"
+        master_file = requests.get(master_file)
+        monster_data = master_file.json()["pokemon"]
+        with open(pa_root + "/data/pokemon_data.json", 'w') as f:
```
billyjbryant, line 15: Ok, I see we are caching the contents of the remote file locally, but there is no error handling here, IE, if the remote file is missing or becomes empty how do we prevent it from overwriting the local cached copy?

billyjbryant: Also, why is this happening in pvp, this should probably have its own utility file that is loaded on PA start.

billyjbryant (changes requested):

tools/pvp.py
Pontiky: As I quickly said in the description above, this file containing the data I used to update PA is from [the same source as the one used by PoracleJS](https://github.com/KartulUdus/PoracleJS/blob/0f79cbde767383bdb9647991c244e09ec569449d/src/util/generateData.js#L23). They just use a different template.

tools/pvp.py
Pontiky: > Ok, I see we are caching the contents of the remote file locally, but there is no error handling here, IE, if the remote file is missing or becomes empty how do we prevent it from overwriting the local cached copy?

The data will be erased only if the pokemon_data is missing or the local hash `data/.data_version` is different than the hash of the remote file. There is no reason why the file could be empty. If this is about connection issues, PA couldn't start sending the alarms anyway even with the local copy.


> Also, why is this happening in pvp, this should probably have its own utility file that is loaded on PA start.

`pvp.py` is an independent tool, which means this is not used by PA and doesn't include any of its modules to work. You could run `python3 tools/pvp.py` if you have the local `pokemon_data.json`.

It is used to calculate PvP stat products of Pokemons by getting their base stats. Before this PR, it used the `base_stats.json` file always available in `data/` folder locally. With this PR who fetch `pokemon_data.json` remotely, there is no way to know if the data is available in local `data/` folder when you run `python3 tools/pvp.py` as it won't start PA. 

Same reason for including [the download of the pokemon data when running the unit tests](https://github.com/pokealarm/pokealarm/pull/802/files#diff-e314724a0aa4e95190e5c3a2443ee29c11ff49771675ed554fd46ba14176a275R14).

Pontiky:
> Please move all locale changes into a separate PR

The locale changes were introduced in this PR to support the new data format. But this is not really a problem if they are added in a different PR as the default form names are the ones from `data/pokemon_data.json`

tools/pvp.py
billyjbryant: > The data will be erased only if the pokemon_data is missing or the local hash data/.data_version is different than the hash of the remote file. There is no reason why the file could be empty. If this is about connection issues, PA couldn't start sending the alarms anyway even with the local copy.

Let's suppose a rogue commit wipes or deletes the remote file. How are we going to deal with that dependency. The hash will be different but the remote file is no longer trustworthy. We need a way to fall-back on previously cached data so that PA's dependency is not impacted.

billyjbryant (approved):

PR #803, opened 2022-02-13
Fix min cp

weizh888:
## Type of Change

## Motivation and Context

The min CP is 10, but based on the CP calculation formula, it gives CP less than 10.

## How Has This Been Tested?

Self test.

## Wiki Update

### DIFF
--- a/PokeAlarm/Utilities/PvpUtils.py
+++ b/PokeAlarm/Utilities/PvpUtils.py
@@ -12,7 +12,7 @@ def calculate_cp(monster, atk, de, sta, lvl):
     cp = ((base_stats["attack"] + atk) * sqrt(base_stats["defense"] + de) *
           sqrt(base_stats["stamina"] + sta) * (multipliers[str(lvl)]**2)
           / 10)
-    return int(cp)
+    return int(cp) if int(cp) > 10 else 10
 
 
 def max_cp(monster):


PokeAlarm/Utilities/PvpUtils.py
```diff
@@ -12,7 +12,7 @@ def calculate_cp(monster, atk, de, sta, lvl):
     cp = ((base_stats["attack"] + atk) * sqrt(base_stats["defense"] + de) *
           sqrt(base_stats["stamina"] + sta) * (multipliers[str(lvl)]**2)
           / 10)
-    return int(cp)
+    return int(cp) if int(cp) > 10 else 10
```
billyjbryant, line 15: Seems like we should be determining what can cause the formula above to result in a CP < 10 as 10 is the base CP for a large majority of Mon. Is the formula wrong or does Niantic abandon the formula for some CP 10 Mon?

weizh888: The cp formula is correct, but there's a min cp, especially for level 1 mons. Based on gamepress: `If CP would be less than 10, it is displayed as 10 instead`

billyjbryant (approved):

billyjbryant (changes requested):
@weizh888 After #802 I am not certain this change is needed. Can you please rebase and resolve the merge conflicts if it is still needed then re-submit?

Pontiky:
> @weizh888 After #802 I am not certain this change is needed. Can you please rebase and resolve the merge conflicts if it is still needed then re-submit?

After #802, `calculate_cp()` function has been moved to `Utils` module at [this line](https://github.com/pokealarm/pokealarm/blob/81283b50e60f5e3e12cac7b0d466fd00afcf277c/PokeAlarm/Utils.py#L463).

This fix is still needed.

PR #804, opened 2022-02-15
Fix evolution cost calculation failing in some cases

Pontiky:
## Description

Mitigate `None` values when calculating the evolution cost of a pokemon.

## Type of Change

## Motivation and Context

Fix rare cases where the evolution cost is missing.

## How Has This Been Tested?

Same tests as <https://github.com/pokealarm/pokealarm/pull/794>

## Wiki Update


billyjbryant (changes requested):
This only has two files changed and a net of +1 lines but there are 12 commits. What happened here? Can you squash your commits please?

billyjbryant (approved):
Going to Squash and Merge

Pontiky:
> This only has two files changed and a net of +1 lines but there are 12 commits. What happened here? Can you squash your commits please?

All the commits (except the 3 lasts) came from #794 since I made this patch from this PR.

PR #805, opened 2022-02-17
Fix crash on PvP calculations

Pontiky:
## Description

Fix a bug where sometimes PA could crash being stuck in a `while` loop.

## Type of Change

## Motivation and Context

Since the last commit, PA sometimes crashes and is unable to stop without killing the terminal. It looks like the cost calculation functions for PvP could infinitely loop in a `while` function. Some restrictions have been added to these `while` to avoid infinite loop.

## How Has This Been Tested?

The freezes happened every times during the first seconds or minutes.
With this patch, I tested it during a few hours and PA didn't freeze again.

## Wiki Update

### DIFF
--- a/PokeAlarm/Utilities/PvpUtils.py
+++ b/PokeAlarm/Utilities/PvpUtils.py
@@ -69,14 +69,14 @@ def calculate_candy_cost(start_level, target_level, evo_candy_cost=0):
     candy_cost = evo_candy_cost
     xl_candy_cost = 0
 
-    while tmp_level != target_level and tmp_level < 40:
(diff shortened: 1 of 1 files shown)


billyjbryant (approved):

billyjbryant (commented):
This needs to be fully tested and vetted

billyjbryant:
I cannot revoke my approval on this PR, I am closing it instead. #806 will revert #794 and #804. Please resubmit these changes once tested as a new PR

billyjbryant:
This PR is confirmed to fix the issue originally introduced in #794 so merged

PR #813, opened 2022-03-02
Fetch move data automatically

### LINKED ISSUE #770: Fetch Base Stats and Move Info

Pontiky:
## Description

Following PR #802 who brought automatic pokemon data update, this PR introduces the automatic move data update and will close #770.
The local `move_info.json` file is now deprecated and has been removed.
To replace it, two new files `fast_moves.json` and `charged_moves.json` are fetched from [Pogo API](https://pogoapi.net/documentation/).

## Type of Change

## Motivation and Context

The `move_info.json` file were behind compared to the current game status since a few months.
[All the moves starting from ID 292 are wrong and not updated.](https://github.com/pokealarm/pokealarm/blob/d686ac8d972d0e3a601830cac865714fac092d65/data/move_info.json#L1304) They have the same damage, DPS, duration and energy. Moreover, the last move added is [move ID 320](https://github.com/pokealarm/pokealarm/blob/d686ac8d972d0e3a601830cac865714fac092d65/data/move_info.json#L1500) while the last one added into the game is [move ID 364](https://pogoapi.net/api/v1/charged_moves.json).

Like the Pokemon data, this update brings the automation of move updates to stop from being behind.
The functions to get move informations have been updated to follow the new data format, some file checks have been added and the fetch functions have been changed to support multiple file updates.

## How Has This Been Tested?

Tested on a live PA.
I also did some tests to compare the old and new output of the move functions (I only wrote 3 examples below):

Old output:
| Move ID  | Type ID | Damage | DPS | Duration | Energy |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| 14 | 1 | 150 | 39.47 | 3800 | 100 |
| 291 | 1 | 5 | 3.8 | 1300 | 20 |
| 315 | 8 | 5 | 3.8 | 1300 | 20 |

New output:
| Move ID  | Type ID | Damage | DPS | Duration | Energy |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| 14 | 1 | 150 | 39.47 | 3800 | 100 |
| 291  | 1 | 5 | 3.85 | 1300 | 20 |
| 315  | 8 | 80 | 28.57 | 2800 | 50 |

The old and new output are the same except for moves ID greater than 292 as they are not up to date in local PA repo.
The output of these function has not been changed. So we shouldn't except errors for functions which call them and use them.
Source to verify move outputs: <https://bulbapedia.bulbagarden.net/wiki/List_of_moves_(GO)>

## Wiki Update


CarlosVMX1:
Had been using this PR for 5 days, all works good and had checked some moves in my live setup and the output looks correct based on the bulbapedia site.

billyjbryant (approved):
Has been tested by multiple users to be functional.

PR #814, opened 2022-03-02
Added virtualenv doc and updated scanner doc

### LINKED ISSUE #715: Python3 Virtualenv Docs
## Description
When python 3 support gets added, people will want to know how to use virtual envs so they can use PA in parallel with other projects (such as MAD) that use python 3.

People will ask

Tap a keyboard for a while until it makes sense

Pontiky:
## Description

This PR adds the `virtualenv` doc to isolate PokeAlarm environment from other Python projects.
I also removed all the references to outdated scanners/tools (Monocle and Rocketmap) and added the new ones (MAD and RDM) with some links and tips to configure them correctly with PA.
I finally changed python commands to `python3` and `pip3` as otherwise it might refer to Python 2 for installations with both Python.

## Type of Change

## Motivation and Context

This will close #715.

## How Has This Been Tested?

No test needed

## Wiki Update

- [x] This change is a Wiki update.

### DIFF
--- a/docs/configuration/geofences.rst
+++ b/docs/configuration/geofences.rst
@@ -89,7 +89,7 @@ Differences from scanners
 If you are using the same Geofences for both the scanner and PokeAlarm, it's
 possible that you may see an event get rejected by PokeAlarm. This is because
 the scanner and PA could use Geofences differently. The scanners use them to
-rescrit work movement, but PA uses them to restrict events. As a result,
+restrict work movement but PA uses them to restrict events. As a result,
 the scanners could occasionally send an Event that PA will reject.
 If this is a problem, you can either increase the size of your PA geofences,
 or remove them all together.
--- a/docs/configuration/geofences.rst
+++ b/docs/configuration/geofences.rst
@@ -83,12 +83,13 @@ every Geofence attached to a Manager:
 Notes
 -------------------------------------
 
-Differences from Rocketmap
+Differences from scanners
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you are using the same Geofences for both Rocketmap and PokeAlarm, it's
+If you are using the same Geofences for both the scanner and PokeAlarm, it's
 possible that you may see an event get rejected by PokeAlarm. This is because
-RM and PA use Geofences differently. RM uses them to rescrit work movement, but
-PA uses them to restrict events. As a result, RM will occasionally send
-an Event that PA will reject. If this is a problem, you can either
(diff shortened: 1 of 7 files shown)


docs/configuration/geofences.rst
```diff
@@ -83,12 +83,13 @@ every Geofence attached to a Manager:
...
-If you are using the same Geofences for both Rocketmap and PokeAlarm, it's
+If you are using the same Geofences for both the scanner and PokeAlarm, it's
 possible that you may see an event get rejected by PokeAlarm. This is because
-RM and PA use Geofences differently. RM uses them to rescrit work movement, but
-PA uses them to restrict events. As a result, RM will occasionally send
-an Event that PA will reject. If this is a problem, you can either
-increase the size of your PA geofences, or remove them all together.
\ No newline at end of file
+the scanner and PA could use Geofences differently. The scanners use them to
+rescrit work movement, but PA uses them to restrict events. As a result,
```
LegitDongo: typo

docs/getting-started/installation.rst
```diff
@@ -66,15 +74,16 @@ Running
...
-   - **For RocketMap:**
-     Configure the webhook types to send info as explained in
-     `RocketMap wiki <https://rocketmap.readthedocs.io/en/develop/extras/webhooks.html>`_.
+   - **For RealDeviceMap:**
+     Configure the webhook URL to send info as explained in
+     `RealDeviceMap wiki <https://realdevicemap.readthedocs.io/en/latest/realdevicemap/dashboard/settings.html#webhook>`_.
 
    Finally, start your scanner. If everything is set up correctly, PA will start
-   logging information about what it is doing to the console.
+   logging informations about what it is doing to the console.
```
LegitDongo: `information` was correct

LegitDongo (changes requested):
TIL that default python in newest ubuntu is `python3`, along with `pip3`

I would have assumed that it would still default to `python` and `pip`, but this isn't the case, these commands are not available (I'm guessing since 2.7 isn't there)


Got a few small typos

Pontiky:
> TIL that default python in newest ubuntu is `python3`, along with `pip3`
> 
> I would have assumed that it would still default to `python` and `pip`, but this isn't the case, these commands are not available (I'm guessing since 2.7 isn't there)

That's it but if you are using `virtualenv` by specifying python3 at creation, you can use both `pip` and `pip3` in this env as the first one will be an alias of the second one. Same with `python` command.

LegitDongo (approved):
LGTM

PR #824, opened 2022-03-31
Add signed GMaps requests for static maps URL

Pontiky:
## Description

This PR adds the ability to generate signed requests when generation the static map URL. I also added documentation to explain how to set it up correctly.

## Type of Change

## Motivation and Context

**The static map of GMaps is currently suffering from a security issue** as it exposes the GMaps API key in static image URLs and anybody who has this key can make his own requests for his application or can even spam requests.

To prevent this, there is something called [digital signature](https://developers.google.com/maps/documentation/maps-static/digital-signature#sample-code-for-url-signing) which allow to generate a signature using:
- the static image URL (which contains the API key)
- a signing secret string which stay secret and never exposed in any way

Once the signing secret is set up, by configuring the API key, we can only allow digitally signed requests and so, even if someone know this API key, it will be useless as it is if the signature cannot be generated to authenticate the request.

By knowing this, this PR introduce the ability to generate signed requests to improve potential malicious uses of your API key.  
**This feature is a non-breaking change for those who will not change their config.** To enable the signed requests, a new parameter in config.ini named `gmaps-signing-secret` has been added and empty by default. If it's not provided, the behaviour will stay the same as before given we can't sign without this secret string. If the secret is provided with the API key, the next requests will be automatically signed.

## How Has This Been Tested?

Tested on a live configuration and all worked like a charm.

## Wiki Update

Added in this PR.

### DIFF
--- a/PokeAlarm/Alarms/Discord/DiscordAlarm.py
+++ b/PokeAlarm/Alarms/Discord/DiscordAlarm.py
@@ -5,8 +5,9 @@ import requests
 
 # Local Imports
 from PokeAlarm.Alarms import Alarm
-from PokeAlarm.Utils import parse_boolean, get_static_map_url, \
-    reject_leftover_parameters, require_and_remove_key, get_image_url
+from PokeAlarm.Utils import parse_boolean, get_gmaps_static_url, \
+    reject_leftover_parameters, require_and_remove_key, get_image_url, \
+    sign_gmaps_static_url
 
 try_sending = Alarm.try_sending
 replace = Alarm.replace
@@ -115,7 +116,8 @@ class DiscordAlarm(Alarm):
     }
 
     # Gather settings and create alarm
-    def __init__(self, mgr, settings, max_attempts, static_map_key):
+    def __init__(self, mgr, settings, max_attempts, static_map_key,
+                 signing_secret_key):
         self._log = mgr.get_child_logger("alarms")
         # Required Parameters
         self.__webhook_url = require_and_remove_key(
@@ -131,6 +133,7 @@ class DiscordAlarm(Alarm):
         self.__avatar_url = settings.pop('avatar_url', "")
         self.__map = settings.pop('map', {})
         self.__static_map_key = static_map_key
+        self.__signing_secret_key = signing_secret_key
         self.__timestamp = settings.pop('footer_timestamp', False)
 
         # Set Alert Parameters
@@ -201,7 +204,7 @@ class DiscordAlarm(Alarm):
             'body': settings.pop('body', default['body']),
             'fields': settings.pop('fields', []),
             'map': map if isinstance(map, str) else
-            get_static_map_url(map, self.__static_map_key)
+            get_gmaps_static_url(map, self.__static_map_key)
         }
 
         timestamp = settings.pop('footer_timestamp', self.__timestamp)
@@ -236,12 +239,15 @@ class DiscordAlarm(Alarm):
                     'lat': info['lat'],
                     'lng': info['lng']
                 }
+                static_map_url = replace(alert['map'],
+                                         coords if not
+                                         isinstance(alert['map'], str)
+                                         else info)
+                if self.__signing_secret_key is not None:
+                    static_map_url = sign_gmaps_static_url(
+                        static_map_url, self.__signing_secret_key)
                 payload['embeds'][0]['image'] = {
-                    'url':
-                        replace(alert['map'],
-                                coords if not
-                                isinstance(alert['map'], str)
-                                else info)
+                    'url': static_map_url
                 }
 
             if 'timestamp' in alert:
--- a/PokeAlarm/Alarms/Discord/DiscordAlarm.py
+++ b/PokeAlarm/Alarms/Discord/DiscordAlarm.py
@@ -235,17 +235,19 @@ class DiscordAlarm(Alarm):
             }]
 
             if alert['map'] is not None:
-                coords = {
-                    'lat': info['lat'],
-                    'lng': info['lng']
-                }
-                static_map_url = replace(alert['map'],
-                                         coords if not
-                                         isinstance(alert['map'], str)
-                                         else info)
-                if self.__signing_secret_key is not None:
-                    static_map_url = sign_gmaps_static_url(
-                        static_map_url, self.__signing_secret_key)
+                static_map_url = ""
+                if not isinstance(alert['map'], str):
+                    coords = {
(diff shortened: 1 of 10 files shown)


PokeAlarm/Alarms/Discord/DiscordAlarm.py
```diff
@@ -233,12 +236,15 @@ def send_alert(self, alert, info):
                     'lat': info['lat'],
                     'lng': info['lng']
                 }
+                static_map_url = replace(alert['map'],
+                                         coords if not
+                                         isinstance(alert['map'], str)
+                                         else info)
+                if self.__signing_secret_key is not None:
```
LegitDongo: This ensures that gmaps always gets used as long as they have a secret key set in their config, regardless of their alarm settings; it might also force-enable gmaps (not sure here)

Best would be to follow the alarm map setting first, then gmaps if it's not set

PokeAlarm/Alarms/Slack/SlackAlarm.py
```diff
@@ -167,12 +169,19 @@ def send_alert(self, alert, info):
...
-                replace(alert['map'],
-                        info if isinstance(map, str) else coords)
-        }] if alert['map'] is not None else None
+        attachments = None
+        if alert['map'] is not None:
+            static_map_url = replace(alert['map'],
+                                     info if
+                                     isinstance(map, str)
+                                     else coords)
+            if self.__signing_secret_key is not None:
```
LegitDongo: Same here as in DiscordAlarm

LegitDongo (changes requested):
Mostly looks fine, needs changes so alarm settings get appropriately used

billyjbryant:
@Pontiky have the changes that @LegitDongo requested been made? If so, they haven't been committed here. If there were secondary conversations that did not take place within the confines of this PR to address this, those need to be captured here for tracking.

LegitDongo:
@billyjbryant see the snippet that got modified in DiscordAlarm

It's an issue of alert hierarchy, when the map setting is set on the alarm-level, we wouldn't expect a gmaps static map just because we have a signed key set in our config

I've been out for a while for various things, think you can check it out when you get the time?

I do plan to get to it, but I've gotta fix my local PA setup as well 😄

LegitDongo:
ref: https://discord.com/channels/215181169761714177/229088248146821140/971878304837697586

Pontiky:
> It's an issue of alert hierarchy, when the map setting is set on the alarm-level, we wouldn't expect a gmaps static map just because we have a signed key set in our config

As of now, the code you highlighted is in `if alert['map'] is not None:` condition. That's the one used to define if we set up the minimap whatever if `gmaps-key` parameter is present in `config.ini`.
cf: https://github.com/pokealarm/pokealarm/blob/1301c6b7b3c8e4439975e7a9c673866eb98fbe39/PokeAlarm/Alarms/Discord/DiscordAlarm.py#L234

That's why I put it signing check in this block to keep the priority `gmaps-key` > `gmaps-signing-secret`.
There is no `if self.__static_map_key is not None` in the code, so I didn't use `if self.__signing_secret_key is not None`.
Should I anyway?

LegitDongo (commented):
Finally got around to testing this

with this alarm/filter/rules configuration: https://pasteyour.cf/efavosejus.pl
and *without* a `gmaps-signing-secret` + with a `gmaps-key` I get a successful notification
image


When I add `gmaps-signing-secret: abc` to my config.ini, I get the following exception:
```
2022-08-09 21:32:56,500 [DEBUG][     test1][    alarms] Attempting to send notification to Discord.
  File "src/gevent/greenlet.py", line 716, in gevent._greenlet.Greenlet.run
  File "C:\Users\Legit Dongo\PycharmProjects\PokeAlarm\PokeAlarm\Alarms\Discord\DiscordAlarm.py", line 264, in pokemon_alert
    self.send_alert(self.__monsters, pokemon_info)
  File "C:\Users\Legit Dongo\PycharmProjects\PokeAlarm\PokeAlarm\Alarms\Discord\DiscordAlarm.py", line 244, in send_alert
    static_map_url = sign_gmaps_static_url(
  File "C:\Users\Legit Dongo\PycharmProjects\PokeAlarm\PokeAlarm\Utils.py", line 818, in sign_gmaps_static_url
    decoded_key = base64.urlsafe_b64decode(secret)
  File "C:\Users\Legit Dongo\AppData\Local\Programs\Python\Python38-32\lib\base64.py", line 133, in urlsafe_b64decode
    return b64decode(s)
  File "C:\Users\Legit Dongo\AppData\Local\Programs\Python\Python38-32\lib\base64.py", line 87, in b64decode
    return binascii.a2b_base64(s)
binascii.Error: Incorrect padding
2022-08-10T02:32:56Z <Greenlet "Greenlet-0" at 0x4f3af08: <bound method DiscordAlarm.pokemon_alert of <PokeAlarm.Alarms.Discord.DiscordAlarm.DiscordAlarm object at 0x04F6C1A8>>({'webhook': 'https://discordapp.com/api/webhooks/3)> failed with Error
```
(I expected an exception here since the signing key is clearly invalid)

More importantly, this shows that the alarm-level setting of `"map"` gets entirely ignored when the `gmaps-signing-secret` is even *included* in the config, and the app attempts to sign something even if there's no gmaps URL to use

This is a fundamental flaw, not everybody will use gmaps for every alarm, some will choose a spit-usage configuration

As for your previous comment, including a gmaps-key does not mean that gmaps always gets used, it still chooses the `map` string over the gmaps configuration since it's defined explicitly on the alarm

Pontiky:
Just committed 2 changes:
- The call of signing url function is not automatic anymore when the signing secret is set
- I also added fallbacks when trying to sign the url. If something went wrong, an error is logged and the unsigned url is used instead to not cancel notification.

I tested the following cases with the webhook tester:
- no api key and no signing secret set: the notification passes without the minimap ✔️ 
- no api key with a valid signing secret: : the notification passes without the minimap ✔️
- no api key with an invalid signing secret: : the notification passes without the minimap ✔️
- api key with no signing secret: the notification passes with an unsigned minimap ✔️
- api key with a valid signing secret: the notification passes with an signed minimap ✔️
- api key with an invalid signing secret: the notification passes with an unsigned minimap ✔️

billyjbryant:
@LegitDongo since you recently reviewed this and found flaws, please re-review at your convenience.

LegitDongo (approved):
Checked most of these, just not the signed version

Google requires a billing account to enable it I'm guessing :(

PR #829, opened 2022-04-02
Pin jinja2 and werkzeug package versions to resolve conflicts with Flask

Pontiky:
## Description

Pin Jinja2 Python package to v3.0.3
Pin Werkzeug Python package to v2.0.3

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (would cause existing functionality to change)

## Motivation and Context

As mentioned by #827, Jinja2 > v3.0.3 is incompatible with Flask  v1.x.x (currently pinned to 1.1.1).
Werkzeug > v2.0.3 is also incompatible with Flask  v1.x.x.
Resolves #827

PR #822, opened 2022-03-25
Package update for PokeAlarm doc

Pontiky:
## Description

(DOC ONLY) Use a compatible package version for `itsdangerous` with Python 2

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (would cause existing functionality to change)

## Motivation and Context

Given PokeAlarm doc is still generated using Python 2.7... 🤦, it's not generated and updated anymore since `itsdangerous` was pinned to v2.0.1 (PR #811).
This is a workaround to use a different version of an incompatible package (doc only) waiting to migrate ReadtheDocs to Python 3.

First build log failing: https://readthedocs.org/projects/pa/builds/16188438/
Full build list: https://readthedocs.org/projects/pa/builds/

## How Has This Been Tested?

Tested by forcing an update of requirements and only `itsdangerous` 2.0.1 is requested and downloaded when using Python 3. So all is good!

## Wiki Update

- [ ] This change requires an update to the Wiki.
- [x] This change does not require an update to the Wiki.

PR #831, opened 2022-05-05
Use different package versions for readthedocs builds

Pontiky:
## Description

Use different package versions for readthedocs builds as it's still using Python2 to generate the doc.
Introduced by pinning some packages with #829 (required)
Build logs : https://readthedocs.org/projects/pa/builds/16835297/

## Type of Change

## Motivation and Context

Be able to generate the documentation

## How Has This Been Tested?

Same workaround than #822
Chosen pinned versions are those used when Readthedocs was free to use any version: https://readthedocs.org/api/v2/build/16835205.txt

## Wiki Update

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,9 @@ pytz==2019.3
 portalocker==1.5.2
 requests==2.22.0
 greenlet==0.4.13
-jinja2==3.0.3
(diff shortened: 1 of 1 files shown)


LegitDongo (approved):
Whoops! Broke the docs

PR #835, opened 2022-06-16
Rainy lure module

LegitDongo:
## Description
Add the rainy lure module item and lure type for stops

Note: looks like GH's editor removed some extra whitespace, gonna ignore that

## Type of Change

## Motivation and Context
Doesn't get reported correctly without it

## How Has This Been Tested?
None

## Wiki Update

Includes wiki updates


Pontiky (approved):
Changes are safe and the translations comes from the game locales.

PR #836, opened 2022-06-17
Weather update

### LINKED ISSUE #696: Boosted Weather Filter
## Description
Scanners (particularly MAD) now only sends weather as `boosted_weather`, so we should have some way for people to check if a mon is boosted at least
ref:
https://github.com/Map-A-Droid/MAD/blob/master/webhook/webhookworker.py#L428
https://github.com/PokeAlarm/PokeAlarm/blob/dev/PokeAlarm/Events/MonEvent.py#L46

People want to know if a mon is boosted -- looks like this is mostly for Ditto detection

Adding a filter

## Side note:
`boosted_weather` in the webhook contains the *current weather*, which is an ID for the weather status, *not* a boolean

### LINKED ISSUE #731: Weather boost DTS for invasions
## Description

Ability to either filter invasions by weather boost, or add weather boost status in alarms.

Hunting a perfect shadow pokemon is a lot easier with weather boost

I hope the feature is very similar to raid weather boost in alarms, but I don't know.

### LINKED ISSUE #754: Pokemon caching does not handle weather changes.
## Description
Currently if PokeAlarm receives a webhook for a pokemon that contains different stats, it gets blocked due to the caching system. 

Reports are no longer valid if a pokemon becomes boosted or unboosted by weather. There could be 100% iv pokemon that have been scanned that were previously not; but ignored due to cache.

From what i gather from what i have read RDM might have support for this, MAD should have this also coming soon. Custom methods could also exist.

It would be great to add logic to compare the encounter ID incomming to a stat to verify that the stats have not changed.

Pontiky:
## Description

These changes bring a lot of fixes related to the weather.

- Weather boost conditions are now fully determined by the weather ID received from the scanners as they only send current weather. Weather boost is now automatically calculated and not set from webhook as they don't contain this info. The monster form update still brought the necessary tools to determine if a special monster is boosted or no.

- Cached monsters now take the current weather into consideration. This means if a monster is weather boosted at 8:50 but lose his boost at 9:00 due to weather changes, the stats change of the monster change and the scanner can send again this monster. As its ID doesn't change with the weather, previously it was rejected by PokeAlarm even if its stats changed, now PokeAlarm process it. Fixes https://github.com/pokealarm/pokealarm/issues/754

- Due to weather boost calculation fixes (1st point), weather filters now work correctly for monster events. Fixes https://github.com/pokealarm/pokealarm/issues/696

- A new system has been added : THE USE OF CACHED DATA! This means that PokeAlarm is now able to determine if a raid is weather boosted when it just hatches, same for a monster spawning or a new invasion detected.

- With the recent addition of the DTS and filters for grunt/invasions and the introduction of the use of cache system, it's now possible to filter and know with new DTS if shadow monsters are boosted or no. As a reminder, a boosted monster (normal or shadow) is 4/4/4 guaranteed, which allow players to track godd shadow monsters more easily. Closes https://github.com/pokealarm/pokealarm/issues/731

New DTS for invasions :
- `weather_id`: Weather ID of the shadow monster.
- `weather`: Weather name of the shadow monster.
- `weather_or_empty`: Weather name of the shadow monster, or empty string if unknown.
- `weather_emoji`: Weather emoji of the shadow monster, or empty string if unknown.
- `boosted_weather_id`: Return weather ID if shadow monster is boosted.
- `boosted_weather`: Return weather name if shadow monster is boosted.
- `boosted_weather_or_empty`: Return weather name if shadow monster is boosted, or empty string if unknown.
- `boosted_weather_emoji`: Return weather emoji if shadow monster is boosted, or empty string if unknown.
- `boosted_or_empty`: Return `boosted` if shadow monster is boosted, or empty string if not.

New filters for invasions :
- `weather`: Accepted weather conditions, by id or name.
- `boosted_weather`: Accepted boosted weather conditions, by id or name.
- `is_boosted_weather`: Accepts or denies based on boosted weather conditions.

## Type of Change

## Motivation and Context

Weather was broken and unreliable.

## How Has This Been Tested?

My own branch 'beta' still include this PR and I use it for almost 2 months. Someone also used it to fix these weather issues and reported this worked. I also tested this PR itself during 1 day and nothing wrong to report.

## Wiki Update

### DIFF
--- a/PokeAlarm/Events/BaseEvent.py
+++ b/PokeAlarm/Events/BaseEvent.py
@@ -18,6 +18,10 @@ class BaseEvent(object):
         # Create an id for this event to be recognized as
         self.id = time.time()
 
+    def update_with_cache(self, cache):
+        """ Update event infos using cached data from previous events. """
+        raise NotImplementedError("This is an abstract method.")
+
     def generate_dts(self, locale, timezone, units):
         """ Return a dict with all the DTS for this event. """
         raise NotImplementedError("This is an abstract method.")
--- a/PokeAlarm/Events/EggEvent.py
+++ b/PokeAlarm/Events/EggEvent.py
@@ -63,6 +63,12 @@ class EggEvent(BaseEvent):
         self.geofence = Unknown.REGULAR
         self.custom_dts = {}
 
+    def update_with_cache(self, cache):
+        """ Update event infos using cached data from previous events. """
+
+        # Nothing to update
+        pass
+
     def generate_dts(self, locale, timezone, units):
         """ Return a dict with all the DTS for this event. """
         hatch_time = get_time_as_str(self.hatch_time, timezone)
--- a/PokeAlarm/Events/GruntEvent.py
+++ b/PokeAlarm/Events/GruntEvent.py
@@ -6,7 +6,8 @@ from PokeAlarm import Unknown
 from . import BaseEvent
 from PokeAlarm.Utils import get_gmaps_link, get_applemaps_link, \
     get_waze_link, get_time_as_str, get_seconds_remaining, get_dist_as_str,\
-    get_type_emoji, get_gender_sym
+    get_type_emoji, get_gender_sym, is_weather_boosted, \
+    get_cached_weather_id_from_coord, get_weather_emoji
 from PokeAlarm.Utilities.GruntUtils import get_grunt_gender_id, \
     get_grunt_mon_type_id, get_grunt_reward_mon_id, get_grunt_mon_battle, \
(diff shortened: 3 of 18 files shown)


Misterh8er:
Weather works in this PR for boost emojis on monsters as well as raids. Also adjusts raid CP for weather boost properly now.

TechG3n:
With Python 3.6.9 you need to pin s2cell==1.5.0 and it works just fine
Nothing crashes and <boosted_weather_emoji> is back in mon alerts.

billyjbryant (approved):

PR #837, opened 2022-06-19
PvP current rank on the leaderboard

### LINKED ISSUE #830: PVP Leaderboard Information
## Description
Most people only care about PVP based on it's official ranking according to popular websites

Frequently requested

1. Grab rank from popular websites, cache and refresh on command/periodically
1a. Unsure what this will require, might be impossible without scraping
2. Add filters that only show x top spots
3. Add some nice DTS appropriately

Pontiky:
## Description

New DTS and filters to know if it's worth to display or capture the monster for the PvP leagues.
If the mon rank is unknown (aka very very bad), an infinity symbol ∞ is displayed instead of 0.

New DTS :
- `great_rank`:  The current rank of the monster in great league
- `ultra_rank`:  The current rank of the monster in ultra league

New filters:
- `min_rank_great`: Minimum current rank of the mon for great league (ex: 1)
- `max_rank_great`: Maximum current rank of the mon for great league (ex: 50)
- `min_rank_ultra`: Minimum current rank of the mon for ultra league (ex: 1)
- `max_rank_ultra`: Maximum current rank of the mon for ultra league (ex: 50)

## Type of Change

## Motivation and Context

Useful information to know and requested by some people
Closes https://github.com/pokealarm/pokealarm/issues/830

## How Has This Been Tested?

Tested with a live configuration before to open this PR.
Some special names have also been tested like Mr. Mime, Farfetch'd or regional forms.

## Wiki Update

### DIFF
--- a/PokeAlarm/Utilities/PvpUtils.py
+++ b/PokeAlarm/Utilities/PvpUtils.py
@@ -122,7 +122,7 @@ def get_pvp_info(monster_id, form_id, atk, de, sta, lvl):
     ultra_id = monster_id
     ultra_candy = calculate_candy_cost(lvl, ultra_level)
     ultra_stardust = calculate_stardust_cost(lvl, ultra_level)
-    ultra_rank = get_pvp_rank(monster_id, form_id, 1500)
+    ultra_rank = get_pvp_rank(monster_id, form_id, 2500)
 
     if float(great_level) < lvl:
         great_rating = 0
--- a/PokeAlarm/Utilities/PvpUtils.py
+++ b/PokeAlarm/Utilities/PvpUtils.py
@@ -1,6 +1,8 @@
 import PokeAlarm.Utils as utils
+import PokeAlarm.Locale as locale
 import logging
 import re
+import json
 
 log = logging.getLogger('PvpUtils')
 
@@ -110,6 +112,7 @@ def get_pvp_info(monster_id, form_id, atk, de, sta, lvl):
     great_id = monster_id
     great_candy = calculate_candy_cost(lvl, great_level)
     great_stardust = calculate_stardust_cost(lvl, great_level)
+    great_rank = get_pvp_rank(monster_id, form_id, 1500)
 
     ultra_product, ultra_cp, ultra_level = pokemon_rating(
         2500, monster_id, form_id, atk, de, sta, utils.min_level(
@@ -119,6 +122,7 @@ def get_pvp_info(monster_id, form_id, atk, de, sta, lvl):
     ultra_id = monster_id
     ultra_candy = calculate_candy_cost(lvl, ultra_level)
     ultra_stardust = calculate_stardust_cost(lvl, ultra_level)
+    ultra_rank = get_pvp_rank(monster_id, form_id, 1500)
 
     if float(great_level) < lvl:
         great_rating = 0
@@ -146,6 +150,9 @@ def get_pvp_info(monster_id, form_id, atk, de, sta, lvl):
                 2500, evo_id, evo_form_id),
             utils.max_level(2500, evo_id, evo_form_id))
 
+        evo_great_rank = get_pvp_rank(evo_id, evo_form_id, 1500)
+        evo_ultra_rank = get_pvp_rank(evo_id, evo_form_id, 2500)
+
         evo_great = 100 * (great_product / best_great_product)
         evo_ultra = 100 * (ultra_product / best_ultra_product)
 
@@ -159,6 +166,7 @@ def get_pvp_info(monster_id, form_id, atk, de, sta, lvl):
             great_cp = evo_great_cp
             great_level = evo_great_level
             great_id = evo_id
+            great_rank = evo_great_rank
             evo_candy_cost = calculate_evolution_cost(
                 monster_id, evo_id, evolutions, evolution_costs)
             great_candy = calculate_candy_cost(
@@ -171,6 +179,7 @@ def get_pvp_info(monster_id, form_id, atk, de, sta, lvl):
             ultra_cp = evo_ultra_cp
             ultra_level = evo_ultra_level
             ultra_id = evo_id
+            ultra_rank = evo_ultra_rank
             evo_candy_cost = calculate_evolution_cost(
                 monster_id, evo_id, evolutions, evolution_costs)
             ultra_candy = calculate_candy_cost(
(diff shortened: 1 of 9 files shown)


PokeAlarm/Utilities/PvpUtils.py
```diff
@@ -119,6 +122,7 @@ def get_pvp_info(monster_id, form_id, atk, de, sta, lvl):
     ultra_id = monster_id
     ultra_candy = calculate_candy_cost(lvl, ultra_level)
     ultra_stardust = calculate_stardust_cost(lvl, ultra_level)
+    ultra_rank = get_pvp_rank(monster_id, form_id, 1500)
```
LegitDongo: Is this a mistake? I think 2500 is supposed to be the third argument

PokeAlarm/Utilities/PvpUtils.py
```diff
@@ -179,6 +188,36 @@ def get_pvp_info(monster_id, form_id, atk, de, sta, lvl):
...
+                elif len(mon_and_form_array) == 3:
+                    mon_and_form_array.pop(2)
+                mon_and_form_str = '_'.join(mon_and_form_array)
+                get_pvp_rank.info[f'cp{maxcp}'][mon_and_form_str] = rank_number
+                rank_number += 1
+
+    loc = locale.Locale('en')
+    mon_name = loc.get_english_pokemon_name(monster_id)
+    form_name = loc.get_english_form_name(monster_id, form_id)
+    form_name = form_name.replace("Alola", "Alolan")
```
LegitDongo: Why do we need to replace the name if English is always "Alolan"?

start_pokealarm.py
```diff
@@ -139,7 +145,13 @@ def check_for_update():
                         'data/shiny_data.json')
                         or sig['shiny_possible'] != shiny_possible_sig),
                     'invasions': (not os.path.isfile('data/invasions.json')
-                                  or sig['invasions'] != invasions_sig)
+                                  or sig['invasions'] != invasions_sig),
+                    'rankings': (
+                        not os.path.isfile('data/rankings-1500.json') or
+                        not os.path.isfile('data/rankings-2500.json') or
+                        not os.path.isfile('data/rankings-10000.json') or
```
LegitDongo, line 152: is 10k the max CP for any league?

start_pokealarm.py
```diff
@@ -195,12 +209,15 @@ def check_for_update():
...
-            not os.path.isfile('data/charged_moves.json') or
-            not os.path.isfile('data/shiny_data.json') or
-            not os.path.isfile('data/invasions.json')):
-        log.critical("Missing PokeAlarm data")
+    missing_data_detected = False
+    for dataname in ['pokemon_data', 'fast_moves', 'charged_moves',
+                     'shiny_data', 'invasions', 'rankings-1500',
+                     'rankings-2500', 'rankings-10000']:
+        if not os.path.isfile(f'data/{dataname}.json'):
+            log.critical(f"Missing PokeAlarm data: {dataname}")
```
LegitDongo: maybe be a bit more explicit with this log
`Missing required external PokeAlarm data:`
This way people don't think it's missing internal files

start_pokealarm.py
```diff
@@ -376,11 +393,51 @@ def download_data(sigdiff=None):
                     invasions_fsize < -0.1):  # -10% diff
                 raise Exception(
                     "remote invasions is smaller "
-                    f"({tmp_invasions_fsize} < {invasions_fsize})")
+                    f"({tmp_invasions_fsize} < {invasions_fsize})"
```
LegitDongo: Looks like a typo

LegitDongo (commented):
Mostly questions, small fixes

PokeAlarm/Utilities/PvpUtils.py
Pontiky: Right, it's a mistake

PokeAlarm/Utilities/PvpUtils.py
Pontiky: It's quite particular to explain. [One source](https://raw.githubusercontent.com/pvpoke/pvpoke/master/src/data/rankings/gobattleleague/overall/rankings-1500.json) uses "Alolan" as form name and [the other](https://raw.githubusercontent.com/WatWowMap/Masterfile-Generator/master/master-latest-everything.json) uses "Alola" instead.
The easiest way to get the rank of a pokemon/form is to compare it using its name instead of its ID. So I need to correct this problem when I'm search the rank. Galarian and Hisuian forms don't have this difference.

start_pokealarm.py
Pontiky: 10k cp is just arbitrary and refers to Master league. This number comes from [source file names](https://github.com/pvpoke/pvpoke/tree/master/src/data/rankings/gobattleleague/overall).

start_pokealarm.py
Pontiky: I agree

start_pokealarm.py
Pontiky: Not really, it just looked better ^^

PokeAlarm/Utilities/PvpUtils.py
LegitDongo: I had another look at the ranking file, wish it had dex IDs instead

We should probably use the `speciesId` field from the ranking file instead, it's too easy to come by abnormalities when doing string manipulation on `speciesName` like this.

the `speciesId` field is more predictable since you can just lowercase the mon name + the form name with some underscores
To support this more, special characters seem to just be removed with exception of the `.` in Mr. Mime
Some examples:
```
"speciesId": "muk_alolan",
"speciesName": "Muk (Alolan)",
"rating": 622,
...
"speciesId": "sirfetchd",
"speciesName": "Sirfetch'd",
"rating": 690,
...
"speciesId": "mr_mime",
"speciesName": "Mr. Mime",
"rating": 335,
```
I know some other monsters have special characters in their name (ex `é`), didn't look up any of those; would you please find some and see how they work in this file?

I'd imagine this would be very easy with some kind of [slugify](https://github.com/un33k/python-slugify) ability here, and just change the seperator to `_`

Pontiky: The problem when you use `speciesId` is that you are unable to know if the monster has a special form whereas the parenthesis in `speciesName` bring you this information. When there is no parenthesis, I set the form name to `Normal` to match with the given monster. In our data, everything which doesn't have a special form is set to Normal.

The truth is there is no easier/clean way to proceed. We have a mon id and a form id in input. If they used them in rankings datas, it would have greatly improved the ease of programming this function.

billyjbryant (changes requested):
Please rebase

PokeAlarm/Utilities/PvpUtils.py
Pontiky: I changed the way the rank is calculated to use `speciesId` from the rankings file and `proto` from the masterfile. Checking some special mon names, this is a more robust way to use both fields.

Checked some random monsters and special ones like Mr. Mime and Sirfetch'd.

LegitDongo (approved):
Looks fine now, ready to merge @billyjbryant

billyjbryant (approved):

PR #840, opened 2022-06-26
Ignore unreleased generations when listing mon evolutions

### LINKED ISSUE #839: Teddiursa (216) evolution Ursaking (901) yields TypeError
## Troubleshooting Checklist
- [x] My scanner install is at the most recent version.
- [x] I've already checked the [FAQ](https://pa.readthedocs.io/en/dev/getting-started/faq.html) and the [Wiki](https://pa.readthedocs.io/en/dev/) already.

## Problem Description
When pokemon_id 216 is encountered the following error occurs:

Stack trace:
  File "PokeAlarm/Events/__init__.py", line 23, in event_factory
    return MonEvent(message)
  File "PokeAlarm/Events/MonEvent.py", line 79, in __init__
    PvpUtils.get_pvp_info(
  File "PokeAlarm/Utilities/PvpUtils.py", line 149, in get_pvp_info
    evo_great = 100 * (great_product / best_great_product)

Thats because get_evolutions returns 901:
import PokeAlarm.Utils as utils
>>> utils.get_evolutions(216)
[217, 901]

And get_best_great_product returns no data for that evolution:
get_best_great_product(901, 0) => None

## Environment
Operating System: linux
Python Version: 3.8.10
PokeAlarm Version: 8d7fe96cdecd8164296882fdb61939c28f6b36bb
Scanner used: MAD

Pontiky:
## Description

The [remote data file](https://raw.githubusercontent.com/WatWowMap/Masterfile-Generator/master/master-latest-everything.json) used by PokeAlarm added evolutions from the new 2022 generation while there is no data for these monsters in the game.

The function listing evolutions and evolution costs were affected. They now ignore unreleased generations to avoid unknown values.

These pokemons have an evolution from the new 2022 gen and caused crashes:

```
Scyther 123 -> 900
Qwilfish 211 -> 904
Sneasel 215 -> 903
Ursaring 217 -> 901
Stantler 234 -> 899
Basculin 550 -> 902
```

## Type of Change

## Motivation and Context

Fixes https://github.com/pokealarm/pokealarm/issues/839
More information can be found on this issue.

## How Has This Been Tested?

The monsters above now only return current released generations.

Example:
Previously, Teddiursa returned [217, 901] (Ursaring and Ursaking) with a candy cost respectively set to [50, 0].
Now, this mon returns [217] (Ursaring) with a candy cost set to [50].

## Wiki Update

### DIFF
--- a/PokeAlarm/Utils.py
+++ b/PokeAlarm/Utils.py
@@ -330,37 +330,40 @@ def get_evolution_chain(j, id_, form_id_, a=None, evolution_details=None):
             pass
         else:
             for evo_id in j[id_].get('forms').get(form_id_).get('evolutions'):
-                if evolution_details:
-                    evo_form_id = j[id_].get('forms').get(form_id_).get(
-                        'evolutions').get(evo_id).get('form')
-                    a.append(f"{evo_id}_{evo_form_id}")
-                else:
-                    a.append(int(evo_id))
-                get_evolution_chain(
-                    j,
-                    str(j[id_].get('forms').get(form_id_).get(
-                        'evolutions').get(evo_id).get('pokemon')),
-                    str(j[id_].get('forms').get(form_id_).get(
-                        'evolutions').get(evo_id).get('form')),
-                    a,
-                    evolution_details)
+                if int(evo_id) <= 898:  # block unreleased generations
+                    if evolution_details:
+                        evo_form_id = j[id_].get('forms').get(form_id_).get(
+                            'evolutions').get(evo_id).get('form')
+                        a.append(f"{evo_id}_{evo_form_id}")
+                    else:
+                        a.append(int(evo_id))
+                    get_evolution_chain(
+                        j,
+                        str(j[id_].get('forms').get(form_id_).get(
+                            'evolutions').get(evo_id).get('pokemon')),
+                        str(j[id_].get('forms').get(form_id_).get(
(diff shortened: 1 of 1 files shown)


to-fr:
Confirming that Teddiursa doesn't cause TypeErrors any more.

billyjbryant:
@Pontiky what is the plan to fix this when these types are released into the game?

Pontiky:
> @Pontiky what is the plan to fix this when these types are released into the game?

@billyjbryant The crashes happen because the [remote file](https://raw.githubusercontent.com/WatWowMap/Masterfile-Generator/master/master-latest-everything.json) updated the evolution fields with the new gen but there is still any data from the game about it: no base stats so no stat products for PvP, no evolution cost, no form IDs, ... All these informations aren't available so the remote file we use don't have them.

2 solutions:
- manually block all the monster IDs greater than the last mon of the last released generation (what I used in this PR)
- automatically block all the monsters IDs undefined in the remote data file, which could lead in a scenario where the new mons are registered in the json but have incomplete data (ex: base stats, Evo cost...).

I would prefer to manually block the new gen until we are sure it will not create any crash when calculating informations.

So to answer to your question, we would have to make a new PR request to increase the last supported mon ID when the new gen will be released.

billyjbryant (approved):

PR #844, opened 2022-07-07
Add support of Python 3.9 and 3.10

### LINKED ISSUE #838: Python 3.9-3.10 support
## Problem Description
PA doesn't support Python 3.9 & 3.10, nor has it been tested

From what I can tell 3.10 does not work at all because of how modules get imported

OS differences should also be checked with this

## Environment
Operating System: Any (unsure if this makes a difference)
Python Version: 3.9 & 3.10
PokeAlarm Version: latest

## Log Output
3.10 test:
```
  File "/home/x/PokeAlarm/start_pokealarm.py", line 663, in <module>
    start_server()
  File "/home/x/PokeAlarm/start_pokealarm.py", line 339, in start_server
    parse_settings(os.path.abspath(os.path.dirname(file)))
  File "/home/x/PokeAlarm/start_pokealarm.py", line 638, in parse_settings
    signal(signal.SIGINT, exit_gracefully)
TypeError: 'module' object is not callable
```

Pontiky:
## Description

Add Python 3.9 and 3.10 support

## Type of Change

**PLEASE UPGRADE YOUR PACKAGES AFTER THIS COMMIT:**
`pip install -r  requirements.txt --upgrade`

## Motivation and Context

Python 3.10 is shipped as a default with Ubuntu 22.04.
Python 3.9 is also very common now.
Closes #838

## How Has This Been Tested?

Tested using a fresh config of Python 3.9.5 and Python 3.10.5
Python 3.7 and 3.8 are still supported with the package upgrades.

## Wiki Update

### DIFF
--- a/start_pokealarm.py
+++ b/start_pokealarm.py
@@ -3,7 +3,7 @@
 # Monkey Patch to allow Gevent's Concurrency
 from datetime import timedelta, datetime
 
-from gevent import monkey
+from gevent import monkey, signal_handler
 monkey.patch_all()
 
 # Standard Library Imports
@@ -684,8 +684,8 @@ def parse_settings(root_path):
         managers[m_name].start()
 
     # Set up signal handlers for graceful exit
-    signal(signal.SIGINT, exit_gracefully)
-    signal(signal.SIGTERM, exit_gracefully)
+    signal_handler(signal.SIGINT, exit_gracefully, signal.SIGINT, None)
+    signal_handler(signal.SIGTERM, exit_gracefully, signal.SIGTERM, None)
 
 
 # Because lists are dumb and don't have a failsafe get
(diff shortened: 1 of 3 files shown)


start_pokealarm.py
```diff
@@ -684,16 +684,16 @@ def parse_settings(root_path):
...
 
 
 # Because lists are dumb and don't have a failsafe get
 def get_from_list(arg, i, default):
     return arg[i] if len(arg) > 1 else default
 
 
-def exit_gracefully():
+def exit_gracefully(signum, frame):
     log.info("PokeAlarm is closing down!")
```
billyjbryant, line 698: maybe add a debug log entry here as to which SigNum we are responding to since we have it available now.

billyjbryant (changes requested):

start_pokealarm.py
Pontiky: Good idea, done!

billyjbryant (approved):

PR #849, opened 2022-07-27
Update monster and form PvP stat products, free new monsters and fix crash on PvP calculations

Pontiky:
## Description

- The stat products data has been updated to include new monsters and monster forms. I generated the new file with `tools/pvp.py` included in PA.
- I also freed monsters ids #899-905 because their stats are now known. PA can now use them on PvP evo calculations and other stuff.
- Fixed crashes with the calculation of PvP stat products where we could divide by 0 or None.

## Type of Change

## Motivation and Context

Some monsters and monster forms were missing.

## How Has This Been Tested?

Local config

## Wiki Update


billyjbryant:
Please indicate how you have tested. Need thorough testing as the last PR you submitted caused failures for multiple users.

TechG3n:
Works good for me

Pontiky:
> Please indicate how you have tested. Need thorough testing as the last PR you submitted caused failures for multiple users.

I added additionnal checks when we have a divisor to avoid 0 or None values
```python
great_rating = 0 if best_great_product == 0 else \
        100 * (great_product / best_great_product)
```

I also checked if pokemons with ID between 898 and 905 have their stats (atk, def, sta) filled and that's the case.
<https://raw.githubusercontent.com/WatWowMap/Masterfile-Generator/master/master-latest-everything.json> and search for `Enamorus`.

I also added  default returned values for `get_best_great_product` and `get_best_ultra_product` to stop from receiving None.
If the combination of the pokemon id and form id is not found, it will try to only use the pokemon id with the default form id (=0). And if nothing has been found, it will return 0 as a fallback which will be caught by the above check of the stat product received.
```python
return (get_best_great_product.info.get(f"{pokemon_id}_{form_id}",
            get_best_great_product.info.get(f"{pokemon_id}_0", 0)))
```

I also used the same condition to filter evolution cost chain (previously `if cost !=0`, now `if int(evo_id) <= 905`) and evolution id chain. It will ensure that the chains will be filtered in the same way.

Finally, I updated the `stat_products.json` using the `pvp.py` inside pokealarm (`python tools/pvp.py`). It took a few dozens of minutes to generate the new file.

billyjbryant (approved):

PR #852, opened 2022-10-08
New max CP and costs DTS for monster evolutions

Pontiky:
## Description

1. New DTS:
- `max_perfect_evo_cp` : Final CP after evolved and maxed out a perfect IV monster (monster and raid events)
- `max_evo_cp` : Final CP after evolved and maxed out the monster (monster event)
- `candy_cost_with_evo` : Candy cost to evolve and power up the monster to its max level (monster event)

3. Fixed a bug where current monster level were used to calculate max CP instead of max level

## Type of Change

## Motivation and Context

The current DTS without evolution gave the max CP and cost for the current monster id.
I think it's even better if you know what CP can a monster reach after being fully evolved. When you catch a Slakoth, you prefer to know that once evolved, he reaches 5010 CP rather than know Slakoth's max CP is 1133.

## How Has This Been Tested?

With my live config. The CP and candy/stardust costs calculation functions are still used in PvP.

## Wiki Update

### DIFF
--- a/PokeAlarm/Events/MonEvent.py
+++ b/PokeAlarm/Events/MonEvent.py
@@ -279,7 +279,7 @@ class MonEvent(BaseEvent):
             # Max out
             'max_cp': calculate_cp(self.monster_id, self.form_id,
                                    self.atk_iv, self.def_iv, self.sta_iv,
-                                   self.mon_lvl),
+                                   50),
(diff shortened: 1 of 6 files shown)


LegitDongo (approved):
Everything looks ok, not sure about `MonEvent.py` 213 to 215 code, but I was able to confirm that it works for regular mons and for forms

PR #855, opened 2022-10-29
JSON5 support

LegitDongo:
## Description
Add JSON5 support, which brings some nice features

1. Comments in JSON files
 * JavaScript-style comments (both single and multi-line) are legal; e.g. `// comment here`  or 
 ```js
/*
comment here 
*/
```

2. Object keys may be unquoted if they are legal ECMAScript identifiers
 * e.g. `filter_name: {}` is allowed now, no longer do you need double quotes
3. Objects and arrays may end with trailing commas.
 * no more errors for accidental commas at the end of a list
4. Strings can be single-quoted, and multi-line string literals are allowed.
 * `body: 'test is here'` is allowed, likewise with multi-lines (which is perfect for the body tag 😉 )
5. `.json5` & `.jsonc` files are allowed
6. See [the json5 documentation](https://json5.org/) for more information

### Downside
It's a little slower on startup because of the JSON5 library

## Type of Change

## Motivation and Context
Lots of unneeded errors and complaints with configuration formatting, let's get rid of most of them

## How Has This Been Tested?
Lightly, probably needs tests for multi-line keys, hexadecimal inputs, positive infinity, negative infinity, NaN, numbers with explicit plus signs, etc.

## Wiki Update

### CODE REVIEW

**Comment by Pontiky** [comment]
That's a good change. I just tested it and my filters and alarms files with comments are correctly loaded by PA.

Maybe you could also add `*.jsonc` files to `.gitignore` list as they automatically support comments without reporting errors.

For those who use VSCode, you can also keep your `*.json` files and add this to your `.vscode\settings.json` to disable related json comment errors:
```
    "files.associations": {
        "*.json": "jsonc"
    }
```

Pontiky (approved):

LegitDongo:
Tested a few more things
1. looks like keys with line breaks are fine, log just shows a bunch of spaces between the text
2. infinity(s) and NaN are fine for keys since it just converts it to a string; for values it breaks things

Everything else breaks PA configuration in some way, and this will never be supported in any capacity since the only real use here would be to break PA

To be more explicit, hexadecimal inputs, infinity(s), NaN, and plus-sign numbers will break PA when used for values, this will not be supported officially and you will be mocked if you try to ask for help with this in your configuration 😄

PR #856, opened 2022-10-31
Use weather boosted status (true or false) instead of weather id to cache a monster

Pontiky:
## Description

Informations used to cache a monster are now:
- its encounter id
- its pokedex id (before: its weight)
- its weather boosted status (before: current weather id)

## Type of Change

## Motivation and Context

Reminders:
- PokeAlarm uses a cache system to not ping 10 times the same monster.
- When the weather changes, the IV changes for the monsters which go from weather boosted to unboosted (or the opposite)

Before Weather update #836:
- the weather information was not used to cache a monster
- That means if a newly boosted monster would pass new filters due to IV changes, it would not trigger any alarm because it's considered as the same monster from the cache system.

Before this PR:
- the current weather id was used to cache a monster
- That means the above problem was fixed
- But for monsters with 2 types, they can be boosted by 2 different weathers (= 2 differents weather id). That means if the IV of a monster which triggers an alarm doesn't change when the weather changes (from unboosted to unboosted or from boosted to boosted), it will trigger the same alarm whereas it's the same monster.

Now with this PR:
- The boosted status is used to know if the monster is boosted or not
- That means the above problems are fixed

## How Has This Been Tested?

With `webhook_tester.py` (modified to always use the same encounter id), I generated a Bulbizarre alarm (grass/poison monster boosted by Clear and Cloody weather). I printed the informations used by the cache system to determine if the monster is still know or not (`{mon.enc_id}{mon.weight}{boosted_status}`

```
78785145???1 (Clear weather)
2022-10-31 11:25:50,981 [ INFO][ manager_0][     rules] Monster Bulbizarre passed 1 rule(s) and triggered 1 alarm(s).
78785145???1 (Clear weather)
78785145???1 (Clear weather)
78785145???1 (Cloudy weather)
78785145???1 (Cloudy weather)
78785145???0 (Fog weather)
2022-10-31 11:30:31,215 [ INFO][ manager_0][     rules] Monster Bulbizarre passed 1 rule(s) and triggered 1 alarm(s).
78785145???0 (Fog weather)
78785145???0 (Snow weather)
```

The same weather state only triggers 1 alarm.

## Wiki Update

### DIFF
--- a/PokeAlarm/Manager.py
+++ b/PokeAlarm/Manager.py
@@ -23,7 +23,8 @@ from .LocationServices import GMaps
 from PokeAlarm import Unknown
 from PokeAlarm.Utilities.Logging import ContextFilter, setup_file_handler
 from PokeAlarm.Utilities.GenUtils import parse_bool
-from .Utils import (get_earth_dist, get_path, get_cardinal_dir)
+from .Utils import (get_earth_dist, get_path,
+                    get_cardinal_dir, is_weather_boosted)
 from . import config
 Rule = namedtuple('Rule', ['filter_names', 'alarm_names'])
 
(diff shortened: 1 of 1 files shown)


LegitDongo (approved):
LGTM

PR #859, opened 2023-01-08
Form, Monster and Move names update

Pontiky:
## Description

- Pokemons from ID 899 to 1008 have been added
- Move IDs from 321 to 379 have been added
- New form names since last local update have been added

## Type of Change

## Motivation and Context

Update as of early 2023

## How Has This Been Tested?

Translations come from general wikis and I (native French speaker):
https://bulbapedia.bulbagarden.net/wiki/List_of_French_Pok%C3%A9mon_names
https://bulbapedia.bulbagarden.net/wiki/List_of_German_Pok%C3%A9mon_names
https://www.pokekalos.fr/pokedex/attaques.html

## Wiki Update


LegitDongo:
Why are we using inconsistent characters for this? (`’` vs `'`)
```diff
-    "Pa'u Style": "Pa Style",
+    "Pa’u Style": "Pa'u Style",
```
Sirfetch'd was also changed to use `'`

As for the unicode changes, I'm cool with using lowercase, but we should be consistent.

Pontiky:
> Why are we using inconsistent characters for this? (`’` vs `'`)
> 
> ```diff
> -    "Pa'u Style": "Pa Style",
> +    "Pa’u Style": "Pa'u Style",
> ```
> 
> Sirfetch'd was also changed to use `'`
> 
> As for the unicode changes, I'm cool with using lowercase, but we should be consistent.

The data comes from https://raw.githubusercontent.com/WatWowMap/Masterfile-Generator/master/master-latest-everything.json

`Pa’u Style` was not recognized specifically because it uses a different apostrophe in this file. The "translation" can still uses simple quote `'`.

LegitDongo:
Understood. I didn't realize lookups were being done with that character

PR #863, opened 2023-01-13
add some unittests for get_pvp_info()

sprintnudel:
## Type of Change

## Motivation and Context

This is in preparation for the upcoming pvp calculation optimizations.
Alas unittests never hurt anyways.

## How Has This Been Tested?

Simply run `python3 -m unittest tests/test_get_pvp_info.py`

## Wiki Update


LegitDongo (approved):
Good enough for now 👍

PR #865, opened 2023-01-19
Pvp info speedup

sprintnudel:
## Description

Speed-up execution of get_pvp_info()

## Type of Change

## Motivation and Context

With sufficiently many mon events coming in within a short timespan PokeAlarm tends to loose events. This is because the POSTs time out while the event queue becomes overly full. The slowdown in processing events is due to PvP calculations being done for every mon event. These calculations take too long for them to be finished before more events come in. This can be specifically noted at the beginning of community days and spotlight hours, when all spawnpoints are updated at once.

While the drop of regular expressions use for the calculations in commit bdd516d41f621cdadf5cacd12d5437d08c94e5ab improved the speed significantly, it is still not fast enough here to go unnoticed.

On my rather old AMD Opteron 3280 the test set of 779 mon events, captured at a recent community day, now takes only ~ 250 ms. That is ~ 22 times faster then without these optimizations. Given that the CPU has slower single core performance than with modern ones your numbers may vary.

Measurement was done with cProfile, ignoring the first mon event to not measure ramp-up costs for the in-memory dict caches for the various JSONs in Utils.py.

Before and after:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      778    0.058    0.000    5.277    0.007 PvpUtils.py:29(get_pvp_info)

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      778    0.035    0.000    0.240    0.000 PvpUtils.py:31(get_pvp_info)
```

As a side product of this `tools/pvp.py` now also runs faster. Before and after:
```
real    37m4.672s
user    36m50.594s
sys     0m3.048s

real    5m3.131s
user    5m1.066s
sys     0m0.489s
```

There are several things done to run faster. The most impactful thing of them is doing less calculations. As it turns out python is slow at math. I'll comment on the most prominent changes and explain the rationale below.

## How Has This Been Tested?

Unit tests have been updated to the API changes and run successfully with the old values.

## Wiki Update


sprintnudel:
https://github.com/pokealarm/pokealarm/blob/0aa7f45a4e3e51f672673ee85ac4ad2ba6d53348/PokeAlarm/Utilities/PvpUtils.py#L8-L21
Here's where the calculations are minimized per mon event. The min/max_level() calculations are not needed anymore since `bisect_levels`, in search of the best CP, needs never more than 8 CP calculations. Even better, since most of the wild mons encountered are not evolved about 2/3 in my test sets only need one calculation to determine that the max. CP is below the limit. The `cp_base`, as I called it, is the part of the CP calc that is not level dependent, so we only do the calc once per set of IVs. Also the `multiplier_squares` are new. They are cached CP multipliers ** 2 for each level, to save the repeated calculations of the values.

sprintnudel:
https://github.com/pokealarm/pokealarm/blob/0aa7f45a4e3e51f672673ee85ac4ad2ba6d53348/PokeAlarm/Events/MonEvent.py#L230-L244

`get_pvp_info()` doesn't return strings anymore to safe the processing cost for all of the mons which don't trigger an alarm. Instead it returns (sets of) numbers. That's why the conversion to the old format in now done during the DTS generation stage.

sprintnudel:
https://github.com/pokealarm/pokealarm/blob/0aa7f45a4e3e51f672673ee85ac4ad2ba6d53348/PokeAlarm/Utils.py#L91-L97

All the instances of `hasattr()` calls in `Utils.py` have been replaced by `try except` blocks instead. The `try` part is basically free of additional cost unless an exception occurs, then the cost is quite heavy in comparison. Since in these cases the exception handling in a one-time cost, we're saving in the long run.

sprintnudel:
https://github.com/pokealarm/pokealarm/blob/0aa7f45a4e3e51f672673ee85ac4ad2ba6d53348/PokeAlarm/Utils.py#L354-L355

Instead of using a string`f"{pokemon_id}_{form_id}"` as the dict key a numerical representation is calculated throughout `Utils.py`. Saves the overhead of string handling.

sprintnudel:
https://github.com/pokealarm/pokealarm/blob/0aa7f45a4e3e51f672673ee85ac4ad2ba6d53348/PokeAlarm/Utils.py#L783-L789

Instead of iterating over the costs and summing up the individual values, stardust and candy costs are represented as lists of costs for each level. Then just `sum()`ing the respective slice of these lists is about twice as fast as iterating. Also went back to the original [powerup_costs.json](https://github.com/pokealarm/pokealarm/blob/0aa7f45a4e3e51f672673ee85ac4ad2ba6d53348/data/powerup_costs.json) format, as it is smaller and easier on the eye.

sprintnudel:
https://github.com/pokealarm/pokealarm/blob/0aa7f45a4e3e51f672673ee85ac4ad2ba6d53348/tools/pvp.py#L41-L43
Only the `*_highest_product` values are written now, as the `*_lowest_product` ones were used nowhere and are not found as a byproduct anymore when bisecting.

https://github.com/pokealarm/pokealarm/blob/0aa7f45a4e3e51f672673ee85ac4ad2ba6d53348/tools/pvp.py#L53-L60
`min/max_level()` have been moved here from `Utils.py` since they are not used anywhere else anymore.

https://github.com/pokealarm/pokealarm/blob/0aa7f45a4e3e51f672673ee85ac4ad2ba6d53348/tools/pvp.py#L76-L80
Also got rid of the outer for loop for the levels, since `bisect_levels()` handles this now for each IV combination.

sprintnudel:
That's all folks. Awaiting your comments and speed measurements from running on your hardware. :kissing_heart:

LegitDongo:
While this is a great improvement, it should be noted that this isn't a fix for the problem you described.

This is one of the things we lost when we had to switch away from multiprocessing, and if we are struggling with too much overhead it might be worth looking into separate-process queue options or use-as-you-need calculations based on configuration. (or both)

This is clearly out of scope, just food for thought.

Pontiky (commented):
I read all the code but without testing it for now. Everything is clear and the majority of the changes are welcome.

sprintnudel:
> This is one of the things we lost when we had to switch away from multiprocessing, and if we are struggling with too much overhead it might be worth looking into separate-process queue options or use-as-you-need calculations based on configuration. (or both)

I think multiprocessing is rather wasteful and rather circumvent it whenever possible. Another approach could be to yield after a certain time, so that other coroutines can run.

LegitDongo:
I did a functionality test, and it seems like things work like they should.
The only thing is that I can't test is if the data is accurate because of my lack of scanner.
@Pontiky , you want to give it a test before merge?

Pontiky:
> I did a functionality test, and it seems like things work like they should. The only thing is that I can't test is if the data is accurate because of my lack of scanner. @Pontiky , you want to give it a test before merge?

@LegitDongo I'm currently testing it and everything seems to work fine for the moment.
I would like to test `tools/pvp.py` too. And I'm waiting for an answer from him for my pending review.

sprintnudel:
> And I'm waiting for an answer from him for my pending review.

Oh, cannot find your question, could you point me to it please.

LegitDongo (approved):
LGTM

PR #866, opened 2023-01-24
New checks for missing monster data

Pontiky:
## Description

PA uses a remote file to get pokemon data: https://raw.githubusercontent.com/WatWowMap/Masterfile-Generator/master/master-latest-everything.json

This data has recently been updated with the last monster generation but some required informations are missing like `forms`.
This PR adds additionnal checks to prevent from searching missing data.

## Type of Change

## Motivation and Context

PA didn't work anymore when it auto updated after a restart.

## How Has This Been Tested?

On my own configuration, PA doesn't crash anymore and send back notifications.

## Wiki Update

### DIFF
--- a/PokeAlarm/Utils.py
+++ b/PokeAlarm/Utils.py
@@ -276,8 +276,8 @@ def get_base_stats(pokemon_id, form_id=0):
             f.close()
             for id_ in j:
                 get_base_stats.info[f"{id_}_0"] = j[id_].get("stats")
-                if not (len(j[id_]["forms"]) == 1 and j[id_]["forms"].get("0")):
-                    for form_id_ in j[id_]["forms"]:
+                if not (len(j[id_].get("forms", {})) == 1 and j[id_]["forms"].get("0")):
+                    for form_id_ in j[id_].get("forms", {}):
                         if (
                             j[id_]["forms"][form_id_]["name"] == "Shadow"
                             or j[id_]["forms"][form_id_]["name"] == "Purified"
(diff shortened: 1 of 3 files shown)


LegitDongo:
@Pontiky please throw this in for the `webhook_test.py` as well please
line 428 becomes: `for form_id_ in pokemon_data[id_].get("forms", {}):`

Note: this PR will conflict with #865

Pontiky:
> @Pontiky please throw this in for the `webhook_test.py` as well please line 428 becomes: `for form_id_ in pokemon_data[id_].get("forms", {}):`
> 
> Note: this PR will conflict with #865

@LegitDongo Done!

PokeAlarm/Utils.py
```diff
@@ -429,7 +429,7 @@ def get_evolution_cost_chain(j, id_, form_id_, a=None):
         if j[id_].get("forms").get(form_id_).get("evolutions") is None:
             pass
         else:
-            for evo_id in j[id_].get("forms").get(form_id_).get("evolutions"):
+            for evo_id in j[id_].get("forms", {}).get(form_id_).get("evolutions", {}):
```
sprintnudel, line 432: ~~I think this is still buggy if `"forms"` is {}.
In that case you `{}.get(form_id_)` which is `None` and then in turn `None.get("evolutions", {})` is evaluated.
Will fix in #865, so no need for action here.~~
Edit: nevermind, this branch of `get_evolution_cost_chain()` is only called with guaranteed to exist `form_id_`s

PR #853, opened 2022-10-21
Black Formatting

LegitDongo:
## Description
[Black](https://github.com/psf/black) is an opinionated code formatter

This increases the character limit per line to 88, as well as sets up a standard for formatting

See [the black docs](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length) on why I added extend-ignore for E203 in `.flake8`, this makes it so line limits aren't mega strict on 88 characters

For code in the future, once black is installed, you can simply run `black .` (for current folder) to format any python it finds

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (would cause existing functionality to change)

## Motivation and Context
Nicer line limits and formatting, easier to work with

## How Has This Been Tested?
Locally

## Wiki Update
- [ ] This change requires an update to the Wiki.
- [x] This change does not require an update to the Wiki.

PR #869, opened 2023-01-28
String format optimization

Pontiky:
## Description

This PR removes the old formatting syntaxes and unify them with f-strings.

- `.format()` have been replaced by f-strings (except for logging messages, replaced by `%s`)
- `%s` have been replaced by f-strings (except for logging messages)
- `"" % var` have been replaced by f-strings
- Thanks to #853, a lot of strings can be written on a single line
- \` and `"` literal brackets in strings have been replaced by `'` (except those which refer to json formatting)
- Removed inconsistent spaces and added missing spaces

## Type of Change

## Motivation and Context

F-Strings is an improved and clearer way to format strings brought by Python 3.
They also process strings faster (between 2x and 5x faster than old formating methods).

https://stackoverflow.com/questions/40714555/python-string-formatting-is-more-efficient-than-format-function
https://www.reddit.com/r/Python/comments/pivojb/performance_comparison_of_string_formatting/

## How Has This Been Tested?

PA is running good.

## Wiki Update

### DIFF
--- a/PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
+++ b/PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
@@ -209,7 +209,7 @@ class PushbulletAlarm(Alarm):
         if req_channel is None and channel_tag is not None:
             self._log.error("Unable to find channel. Pushing to all devices instead.")
         else:
-            self._log.debug(f"Setting to channel {channel_tag}.")
+            self._log.debug("Setting to channel %s.", channel_tag)
         return req_channel
 
     # Push a link to the given channel
@@ -231,19 +231,19 @@ class PushbulletAlarm(Alarm):
             "https://api.pushbullet.com/v2/pushes", data=json.dumps(data)
         )
         if res.status_code != requests.codes.ok:
-            self._log.debug(f"PushBullet response was {res.content}")
+            self._log.debug("PushBullet response was %s", res.content)
             raise requests.exceptions.RequestException(
                 f"Response received {res.status_code}, webhook not accepted."
             )
 
-        self._log.debug(f"Notification successful (returned {res.status_code})")
+        self._log.debug("Notification successful (returned %s)", res.status_code)
 
     def update_channels(self):
         response = self._session.get(
             "https://api.pushbullet.com/v2/channels", timeout=30
         )
         if response.ok is not True:
-            self._log.debug(f"Pushbullet channels response was {response.content}")
+            self._log.debug("Pushbullet channels response was %s", response.content)
             raise requests.exceptions.RequestException(
                 f"Response received {response.status_code} channel grabbing not successful"
             )
@@ -251,7 +251,7 @@ class PushbulletAlarm(Alarm):
             self.__channels = response.json()["channels"]
         except KeyError:
             self._log.error("Problem with the PushBullet API")
-            self._log.debug(f"PushBullet /v2/channels response:{response.json()}")
+            self._log.debug("PushBullet /v2/channels response: %s", response.json())
         self._log.debug(
-            f"Detected the following PushBullet channels: {self.__channels}"
+            "Detected the following PushBullet channels: %s", self.__channels
         )
--- a/PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
+++ b/PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
@@ -236,7 +236,7 @@ class PushbulletAlarm(Alarm):
                 f"Response received {res.status_code}, webhook not accepted."
             )
 
-        self._log.debug("Notification successful (returned {res.status_code})")
+        self._log.debug(f"Notification successful (returned {res.status_code})")
 
     def update_channels(self):
         response = self._session.get(
--- a/PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
+++ b/PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
@@ -48,7 +48,7 @@ class PushbulletAlarm(Alarm):
         "raids": {
             "title": "Level <raid_lvl> raid is available against <mon_name>!",
             "url": "<gmaps>",
-            "body": "The raid is available until <24h_raid_end>" " (<raid_time_left>).",
+            "body": "The raid is available until <24h_raid_end> (<raid_time_left>).",
         },
         "weather": {
             "title": "The weather has changed!",
@@ -209,7 +209,7 @@ class PushbulletAlarm(Alarm):
         if req_channel is None and channel_tag is not None:
             self._log.error("Unable to find channel. Pushing to all devices instead.")
         else:
-            self._log.debug("Setting to channel %s." % channel_tag)
+            self._log.debug(f"Setting to channel {channel_tag}.")
         return req_channel
 
     # Push a link to the given channel
@@ -236,7 +236,7 @@ class PushbulletAlarm(Alarm):
                 f"Response received {res.status_code}, webhook not accepted."
             )
 
-        self._log.debug("Notification successful " f"(returned {res.status_code})")
+        self._log.debug("Notification successful (returned {res.status_code})")
 
     def update_channels(self):
         response = self._session.get(
@@ -245,8 +245,7 @@ class PushbulletAlarm(Alarm):
         if response.ok is not True:
             self._log.debug(f"Pushbullet channels response was {response.content}")
             raise requests.exceptions.RequestException(
-                f"Response received {response.status_code}, "
-                "channel grabbing not successful"
+                f"Response received {response.status_code} channel grabbing not successful"
             )
         try:
             self.__channels = response.json()["channels"]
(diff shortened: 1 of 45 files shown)


PokeAlarm/Alarms/Pushbullet/PushBulletAlarm.py
```diff
@@ -236,7 +236,7 @@ def push(self, data):
                 f"Response received {res.status_code}, webhook not accepted."
             )
 
-        self._log.debug("Notification successful " f"(returned {res.status_code})")
+        self._log.debug("Notification successful (returned {res.status_code})")
```
sprintnudel: This needs to be an f-string

PokeAlarm/Alarms/Slack/SlackAlarm.py
```diff
@@ -207,11 +207,7 @@ def send_alert(self, alert, info):
         self.send_message(
             channel=replace(alert["channel"], info),
             username=replace(alert["username"], info),
-            text="<{}|{}> - {}".format(
-                replace(alert["url"], info),
-                replace(alert["title"], info),
-                replace(alert["body"], info),
-            ),
+            text=f'<{replace(alert["url"], info)}|{replace(alert["title"], info)}> - {replace(alert["body"], info)}',
```
sprintnudel: This is not so pretty, maybe have three vars in the f-string instead

PokeAlarm/Alarms/Twitter/TwitterAlarm.py
```diff
@@ -130,16 +130,11 @@ def connect(self):
...
-                "status": (
-                    f"{timestamps[2]} - "
-                    + (
-                        "PokeAlarm has initialized!"
-                        if self.__startup_text == ""
-                        else self.__startup_text
-                    )
-                )
-            }
+            args = {}
```
sprintnudel: This line is not needed, as `args` is guaranteed to be initialized

PokeAlarm/Filters/BaseFilter.py
```diff
@@ -57,12 +57,12 @@ def check_event(self, event):
 
     def reject(self, event, attr_name, value, required):
         """Log the reason for rejecting the Event."""
-        self._log.info("'%s' %s rejected by '%s'", event.name, self._type, self._name)
-        self._log.debug("%s incorrect: (%s to %s)", attr_name, value, required)
+        self._log.info(f"'{event.name}' {self._type} rejected by '{self._name}'")
+        self._log.debug(f"{attr_name} incorrect: ({value} to {required})")
```
sprintnudel: Generally, I think log messages should not be converted to f-strings, but stay as individual arguments like `info(fmt, *args)`. The `logging` package docs ([here](https://docs.python.org/3/howto/logging.html#optimization)) say the actual formatting string is evaluated as late as possible, but with the f-string approach it is evaluated immediately, even when no logging is going to be performed because the log-level is lower. I profiled mon alarm generation yesterday and this reject method showed up pretty high on top:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)      
    59940    0.221    0.000    0.600    0.000 BaseFilter.py:238(__call__)                                                                                  
      999    0.198    0.000    1.022    0.001 MonEvent.py:207(generate_dts)                                                                                
    49950    0.140    0.000    0.292    0.000 BaseFilter.py:58(reject)                       
```                                                              
i.e. it gets called ~50,000 times for generating 999 mon alarms (for my filter config). With the f-strings in place I think it will use way more time.

sprintnudel: Just tried it, times do double roughly:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)                                                                                    
    49950    0.410    0.000    0.572    0.000 BaseFilter.py:58(reject)  
    59940    0.233    0.000    0.896    0.000 BaseFilter.py:238(__call__)                                                                                  
```

Pontiky: You revealed a good point. After some searches, it looks to be actually recommended to use %s formatting instead of F-Strings to not systematically evaluate strings. Especially when the 2/3 of the logs are at debug level and thus thrown away by default.

I made the changes and all above typos.

PokeAlarm/Load.py
```diff
@@ -17,38 +17,35 @@
...
         log.error(
-            "Encountered error while loading Filters: "
-            "{}: {}".format(type(e).__name__, e)
+            "Encountered error while loading Filters: %s: %s", type(e).__name__, e
         )
         log.error(
-            "PokeAlarm was unable to find a filters file "
-            "at {}. Please check that this file exists "
-            "and that PA has read permissions.".format(filepath)
+            "PokeAlarm was unable to find a filters file at {filepath}. "
```
LegitDongo: `filepath` doesn't get passed here and this log string isn't in the `%s` format

PokeAlarm/Locale.py
```diff
@@ -233,41 +233,29 @@ def adjective_placement(self):
...
-            if monster["form"] != 0
-            and self.get_form_name(monster["id"], monster["form"])
-            not in ["Normal", "Normale"]
-            else "",
-            monster=self.get_pokemon_name(monster["id"]),
-        )
+        monster_name = self.get_pokemon_name(monster["id"])
+        form_name = self.get_form_name(monster["id"], monster["form"])
+
+        if monster["form"] != 0 and form_name not in ["Normal", "Normale"]:
```
LegitDongo: Let's change the complexity by changing this block to
```py
if monster["form"] == 0 or form_name in ['Normal", "Normale"]:
  return f"{monster_name}"

if self.adjective_placement():
  return f"{form_name} {monster_name}"

return f"{monster_name} {form_name}"
```

PokeAlarm/Locale.py
```diff
@@ -233,41 +233,29 @@ def adjective_placement(self):
...
         item_name = self.get_item_name(item["id"])
-        reward_template = "{amount} {item}"
-        if not self.adjective_placement():
-            reward_template = "{item} {amount}"
-        return reward_template.format(
-            amount=item["amount"], type=item["type"], item=item_name
-        )
+        if self.adjective_placement():
+            return f'{item["amount"]} {item_name}'
+        else:
```
LegitDongo: `else` is unnecessary here

PokeAlarm/LocationServices/GMaps.py
```diff
@@ -212,40 +210,38 @@ def reverse_geocode(self, latlng, language="en"):
...
+            raise ValueError(f"DM doesn't support mode '{mode}'.")
         # Estimate to about ~1 meter of accuracy
-        origin = "{:.5f},{:.5f}".format(origin[0], origin[1])
-        dest = "{:.5f},{:.5f}".format(dest[0], dest[1])
+        origin = f"{origin[0]:.5f},{origin[1]:.5f}"
+        dest = f"{dest[0]:.5f},{dest[1]:.5f}"
 
         # Check for memoized results
-        key = origin + ":" + dest
+        key = f"{origin} : {dest}"
```
LegitDongo: spaces were added in the string where there wasn't any, was this intentional?

PokeAlarm/Utilities/MonUtils.py
```diff
@@ -53,7 +52,7 @@ def get_move_id(move_name):
             return int(name)  # try as an integer
     except Exception:
         raise ValueError(
-            "Unable to interpret `{}` as a valid " " move name or id.".format(move_name)
+            f"Unable to interpret '{move_name}' as a valid  move name or id."
```
LegitDongo: extra space added before "move"

tests/filters/__init__.py
```diff
@@ -109,17 +100,14 @@ def full_test(self):
...
+                f"pass_val failed check in {test.__name__}: \n{event} passed {filt}",
             )
 
         for val in test.fail_items:
             event = self.gen_event(val)
             self.assertFalse(
                 filt.check_event(event),
-                "fail_val  passed check in {}: \n{} passed {}"
-                "".format(test.__name__, event, filt),
+                f"fail_val  passed check in {test.__name__}: \n{event} passed {filt}",
```
LegitDongo: extra space after fail_val
looks like it was there originally, doubt it was intentional 😄

LegitDongo (changes requested):
Found a few issues

Also, I noticed that you kept the `.format` formatting on `great_pvpoke` and `ultra_pvpoke`
I get why, it would be a very long f-string, just making sure that was intentional.

LegitDongo:
I also gave this a basic test, everything went fine.

PokeAlarm/LocationServices/GMaps.py
Pontiky: Not intentional

Pontiky:
> Found a few issues
> 
> Also, I noticed that you kept the `.format` formatting on `great_pvpoke` and `ultra_pvpoke` I get why, it would be a very long f-string, just making sure that was intentional.

It was intentional because at first sight the f-string would cause them to be more unreadable than they were. But thinking about intermediate variables, I found a proper solution to get ride of these last 2 `format()`. Done with last commit.

LegitDongo (approved):
I'm not a big fan of the new changes to MonEvent, but it does accomplish the goal of this PR
Also, I'm not sure how I would make it any better tbh

PR #872, opened 2023-02-01
optimize Alarm.replace()

sprintnudel:
The loop is very inefficient with the evergrowing number of DTS values. By using a one-pass regex substitution the method is ~ 10x faster.

## Type of Change

## Motivation and Context

Profiling the (Telegram) alarm generation the high number of str.replace() and str.format() calls jumped into my eyes.
While this was no dealbreaker in the sense that it takes too long in a reasonable setup, I still couldn't live with this. With my test json resulting in 999 alarms the cumulative runtime can be brought down by one magnitude using regex:
```
   #ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     #4995    1.128    0.000    1.907    0.000 Alarm.py:77(replace)

   #ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     #4995    0.011    0.000    0.182    0.000 Alarm.py:78(replace)
```

## How Has This Been Tested?

Measurements with cProfile

## Wiki Update

### DIFF
--- a/PokeAlarm/Alarms/Alarm.py
+++ b/PokeAlarm/Alarms/Alarm.py
@@ -1,6 +1,7 @@
 # Standard Library Imports
 import time
 import traceback
+import re
 
 # 3rd Party Imports
 # Local Imports
@@ -79,10 +80,11 @@ class Alarm(object):
         if string is None:
             return None
 
-        s = string
-        for key in pkinfo:
-            s = s.replace(f"<{key}>", str(pkinfo[key]))
-        return s
+        return re.sub(
+            r"<([^<>\n]+)>",
+            lambda m: str(pkinfo.get(m.group(1), f"<{m.group(1)}>")),
+            string,
+        )
 
     @staticmethod
     def pop_type(data, param_name, kind, default=None):


sprintnudel:
Note that this currently conflicts with #869. Which is merged first forces the other to resolve the conflict I guess.

sprintnudel:
Runs on my instance, since this PR was published. No complaints so far.

Pontiky:
The discord webhooks formatted to ping a discord role don't work anymore. Pings are formatted like that: `<@123456789>` and it now simply disappears in the webhook text.

I suspect this PR so we need to deep a bit more to confirm that.

sprintnudel:
> The discord webhooks formatted to ping a discord role don't work anymore. Pings are formatted like that: `<@123456789>` and it now simply disappears in the webhook text.

Should be fixed now.

Pontiky:
> > The discord webhooks formatted to ping a discord role don't work anymore. Pings are formatted like that: `<@123456789>` and it now simply disappears in the webhook text.
> 
> Should be fixed now.

It now works properly 👍

LegitDongo:
This will give some false positives; it won't replace anything, but we can do less work by fixing it
That being said, we might mess up someone's character formatting if we don't fix this
examples:
`<<1>` 
`<<<<<<<<>`
`<3<<>`
this also matches newlines:
```
<
a>
```

`<([^<>\n]+)>` should be an appropriate regex

note: I'm not sure if I tested *every* possibility 😄

LegitDongo (changes requested):
marking this appropriately, see above comment

d12fk:
> `<([^<>\n]+)>` should be an appropriate regex

done

sprintnudel:
> `<([^<>\n]+)>` should be an appropriate regex

done

billyjbryant:
@LegitDongo is your concerns with this PR resolved now?

billyjbryant (approved):

LegitDongo (approved):
haven't tested latest, but seems fine

PR #877, opened 2023-03-21
Golden lure module

Pontiky:
## Description

Golden lure module is the new type of lure.

## Type of Change

## Motivation and Context

Asked by some PA users

## How Has This Been Tested?

Translations have been taken from the official announcement: https://pokemongolive.com/en/post/gocoin/

## Wiki Update


LegitDongo (approved):
There's some scope creep here, but since it's limited to locales and it's removing some fat, I'm cool with that

